PR #4, opened 2021-11-03
Tests for system metrics

mdujava:


testsuite/tests/prometheus/apicast/test_metrics.py
```diff
@@ -58,28 +59,36 @@ def test_metrics_from_target_must_contains_apicast_metrics(expected_metric, metr
...
-        status: int(hit["value"][1]) for status in STATUSES for hit in hits if int(hit["metric"]["status"]) == status}
+        status: int(hit["value"][1]) for status in STATUSES
+        for hit in hits if int(hit["metric"]["status"]) == status and
+        hit["metric"]["container"] == container
+    }
 
 
 @pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-5417")
-def test_apicast_status_metrics(client, prometheus):
+@pytest.mark.parametrize(("client", "apicast"), [("api_client", "apicast-staging"),
```
mganisin, line 77: `apicast` is called `container` elsewhere, it should be called everywhere same (and maybe in this context container seems bit better, however I can't say)

testsuite/tests/prometheus/apicast/test_metrics.py
```diff
@@ -58,28 +59,36 @@ def test_metrics_from_target_must_contains_apicast_metrics(expected_metric, metr
...
     """Reliable gathering of prometheus metrics
 
     Metrics in prometheus appear with some delay, therefore retry is needed
     to ensure expected values are available"""
     hits = prometheus.get_metric("apicast_status")
 
     return {
-        status: int(hit["value"][1]) for status in STATUSES for hit in hits if int(hit["metric"]["status"]) == status}
+        status: int(hit["value"][1]) for status in STATUSES
+        for hit in hits if int(hit["metric"]["status"]) == status and
```
mganisin, line 71: this indentation looks weird, I am surprised pylint doesn't complain (I do), I'd expect one level right (unless linters complain as they are more authoritative than I)
Also `and` at the beginning of line is much better than at the end.

(Btw I don't like this construct at all, however I already accepted it so not complaining here and now)

testsuite/tests/prometheus/system/test_internal_calls.py
```diff
@@ -0,0 +1,118 @@
...
+
+import pytest
+import requests
+
+from packaging.version import Version  # noqa # pylint: disable=unused-import
+from testsuite import TESTED_VERSION, rawobj  # noqa # pylint: disable=unused-import
+from testsuite.prometheus import PROMETHEUS_REFRESH
+
+pytestmark = [
+    # can not be run in parallel
```
mganisin, line 15: this is insufficient, one line comment is great, however it must be explanatory

mganisin (dismissed):

testsuite/tests/prometheus/apicast/test_metrics.py
mdujava: Linters are complaining, It does not look right for me neither. I tried all permutations of one indentation on 1st, 2nd or on both lines at the same time, still not satisfied.

testsuite/tests/prometheus/apicast/test_metrics.py
mdujava: Right, I was in impression that I used getfixturevalue and used the clients, but it's not the case and it's used only as container name, will fix it.

testsuite/tests/prometheus/system/test_internal_calls.py
mdujava: This comment is not accurate in this case, will remove it.

testsuite/tests/prometheus/system/test_internal_calls.py
```diff
@@ -0,0 +1,118 @@
...
+    assert "rails_requests_total" in metrics
+
+    counts_after = extract_call_metrics(prometheus, "rails_requests_total", "backend-listener",
+                                        lambda x: x['metric']['controller'].startswith('admin/api'))
+
+    assert counts_after == counts_before
+
+
+# pylint: disable=unused-argument
+@pytest.mark.flaky
```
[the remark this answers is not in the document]
mdujava: PROMETHEUS_REFRESH is default scrape interval `+1 sec`, so it should always check new data, those failures might be caused by some concurrent usage of the 3scale instance, will test it with dedicated instance.

testsuite/tests/prometheus/system/test_internal_calls.py
```diff
@@ -0,0 +1,118 @@
...
+    counts_before = extract_call_metrics(prometheus, "rails_requests_total", "system-provider",
+                                         lambda x: x['metric']['controller'].startswith('admin/api'))
+
+    service_id = application.service.entity_id
+    app_plan_id = application.entity['plan_id']
+    requests.get(f"{backend_listener_url}/internal/services/{service_id}/plans/{app_plan_id}/usagelimits",
+                 verify=False, headers=auth_headers)
+
+    # prometheus is downloading metrics periodicity, we need to wait for next fetch
+    time.sleep(PROMETHEUS_REFRESH)
```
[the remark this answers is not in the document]
mdujava: in `Prometheus > Targets` there is data of last scrape, I can investigate if we can poll or fetch date of next scrape to wait only necessary time.

pehala (approved):

mganisin (approved):

PR #5, opened 2021-11-04
Use groupified versioning in Templates

pehala:
- Recent versions of `oc` have started throwing warnings about using ungroupified versions (`v1` instead of `template.openshift.io/v1`) when using new-app, they are warnings for now but they might become unsupported later.
- Other usages of such versioning were not causing the warnings to I didn't change them (notably, inside the templates)

### DIFF
--- a/testsuite/resources/apicast.yml
+++ b/testsuite/resources/apicast.yml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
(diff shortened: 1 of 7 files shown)


mganisin:
Sigh... this so pointless change, just because some decided that breaking changes a are good (yes, I agree likely one day former format will stop working)

mganisin (approved):

JaurbanRH (approved):

dhlavac (approved):

PR #6, opened 2021-11-04
Add tests for combination of token scopes

JaurbanRH:
@dhlavac @jsmolar please review

### DIFF
--- a/testsuite/tests/ui/tokens/test_combinations.py
+++ b/testsuite/tests/ui/tokens/test_combinations.py
@@ -1,0 +1,102 @@
+"""
+Tests for combination of token scopes
+"""
+
+import pytest
+
+from testsuite.ui.views.admin.settings.tokens import Scopes, TokenNewView
+from testsuite.utils import blame
+
+
+@pytest.fixture(scope="module")
+def token(custom_admin_login, navigator, request, threescale, permission):
+    """
+    Create token with all scopes
+    """
+    custom_admin_login()
+    new = navigator.navigate(TokenNewView)
+    name = blame(request, "token")
+
+    def _delete():
+        token = [x["name"] == name for x in threescale.access_tokens.list()][0]
+        threescale.access_tokens.delete(token.entity_id)
+
+    request.addfinalizer(_delete)
+    token = new.create(name,
+                       [Scopes.BILLING.value, Scopes.MANAGEMENT.value, Scopes.POLICY.value, Scopes.ANALYTICS.value],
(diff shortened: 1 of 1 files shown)


testsuite/tests/ui/tokens/test_combinations.py
```diff
@@ -0,0 +1,98 @@
...
+from testsuite.ui.views.admin.settings.tokens import Scopes, TokenNewView
+from testsuite.utils import blame
+
+
+@pytest.fixture(scope="module")
+def token(custom_admin_login, navigator, request, threescale, permission):
+    """
+    Create token with all scopes
+    """
+    custom_admin_login()
```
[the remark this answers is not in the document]
JaurbanRH: Because normal login has scope "function"

testsuite/tests/ui/tokens/test_combinations.py
```diff
@@ -0,0 +1,98 @@
...
+    new = navigator.navigate(TokenNewView)
+    name = blame(request, "token")
+    token = new.create(name,
+                       [Scopes.BILLING.value, Scopes.MANAGEMENT.value, Scopes.POLICY.value, Scopes.ANALYTICS.value],
+                       permission[0])
+
+    def _delete():
+        token = list(filter(lambda x: x["name"] == name, threescale.access_tokens.list()))[0]
```
mganisin, line 24: use list comprehension

testsuite/tests/ui/tokens/test_combinations.py
```diff
@@ -0,0 +1,98 @@
...
+def test_read_service(token, api_client):
+    """
+    Request to get list of services should have status code 200
+    """
+
+    response = api_client("GET", '/admin/api/services', token)
+    assert response.status_code == 200
+
+
+def test_create_account_user(account, token, api_client, request, permission):
```
mganisin, line 40: where are these created objects deleted?

mganisin (changes requested):

testsuite/tests/ui/tokens/test_combinations.py
JaurbanRH: account's user is deleted together with account during account fixture cleanup

dhlavac:
@JaurbanRH  what is the difference from current token tests?

JaurbanRH:
After discussion with @dhlavac  we agreed that this PR is redundant and different approach is needed for testing tokens.

PR #8, opened 2021-11-04
Improve performance for on_failed tests

pehala:
* The 503 errors are by default retried, this removes the retrying.
@palmieric WDYT?

### DIFF
--- a/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
+++ b/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
@@ -53,7 +53,7 @@ def test_on_failed_policy(application, status_code):
     Sends request to apicast and check that the returned code
     is the expected one as per `status_code`
     """
-    api_client = application.api_client()
+    api_client = application.api_client(disable_retry_status_list=(503,))
 
     response = api_client.get("/")
     assert response.status_code == status_code
--- a/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
+++ b/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
@@ -2,6 +2,7 @@
 Create a service with a non existent policy in the chain
 and tests that on failed policy returns the correct error code.
 """
+import backoff
 import pytest
 
 from testsuite import rawobj
@@ -48,6 +49,14 @@ def status_code(chain_name, on_failed_configuration) -> int:
     return on_failed_configuration.get("error_status_code", 503)
 
 
+@backoff.on_predicate(backoff.fibo,
+                      lambda response: response.headers.get("server") != "openresty",
+                      max_tries=5)
+def make_request(api_client):
+    """Make request to the product and retry if the response isn't from APIcast """
+    return api_client.get("/")
+
+
 def test_on_failed_policy(application, status_code):
     """
     Sends request to apicast and check that the returned code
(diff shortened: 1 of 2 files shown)


testsuite/tests/apicast/policy/on_failed/test_onfailed.py
```diff
@@ -53,7 +53,7 @@ def test_on_failed_policy(application, status_code):
     Sends request to apicast and check that the returned code
     is the expected one as per `status_code`
     """
-    api_client = application.api_client()
+    api_client = application.api_client(disable_retry_status_list=(503,))
```
mganisin, line 56: hmm, and how to distinguish between setup delay and desired behavior. This seems poisonous

mganisin (changes requested):

testsuite/tests/apicast/policy/on_failed/test_onfailed.py
pehala: That's the problem of the entire test as we are checking only 503 but not its cause. @palmieric Could we set up the maintenance mode to return a different status code so the initial setup 503 is distinguishable?

palmieric:
@pehala I think you have a point here, but I actually would disable the retry on 500 error (the expected return from the policy configuration) and leave untouched the 503 that could happen for other reasons.

testsuite/tests/apicast/policy/on_failed/test_onfailed.py
palmieric: Agree with Marian, you can't distinguish between 503 returned by chance and by actual behavior.

palmieric: @pehala the tests in place are both for:
- default configuration (apicast returns 503 if a policy fails/is skipped)
- custom configuration (apicast returns 500)
We can't bypass the default behavior or a possible configuration is lacking.

pehala: in that case, we cannot be sure if the test passes with the default behaviour regardless of the ignore as it will retry and then still return 503.

How about we first set up service without policies, wait for it to become ready and then add the policy, in that case the route and everything should already be in place and we know that the 503 is from the policy.

Another solution would be checking the content of the response if it contains stuff related to the policy, like some header or seomething.

palmieric (dismissed):
This could be merged

testsuite/tests/apicast/policy/on_failed/test_onfailed.py
mganisin: it is possible to distinguish whether response came from apicast (openresty) or not. Custom backoff based retry on error code **and** "server" header seems to be a choice

mganisin: The question is how to modify it and who should do that (and if reuse this PR) I'll keep the decision on you two :-)

pehala: @mganisin I already changed the PR to backoff on non-apicast answer, could you please re-review it?

testsuite/tests/apicast/policy/on_failed/test_onfailed.py
```diff
@@ -48,12 +49,21 @@ def status_code(chain_name, on_failed_configuration) -> int:
     return on_failed_configuration.get("error_status_code", 503)
 
 
+@backoff.on_predicate(backoff.fibo,
+                      lambda response: "server" not in response.headers or response.headers["server"] != "openresty",
```
mganisin, line 53: what about `response.headers.get()`

mganisin (approved):

palmieric (approved):
LGTM!

mganisin (approved):

PR #14, opened 2021-11-09
Gateway rewrite

pehala:
* Another attempt at gateway rewrite, this time without any builders
  * All parameters are now in the constructors
* Add SettingsParser
  * This class transforms string settings into objects (Like Openshift project and server into OpenShiftClient)
  * Example:
```yaml
openshift:
 project_name: "3scale-alpha-16"
 kind: OpenShiftClient
```


testsuite/gateways/apicast/template.py
```diff
@@ -1,156 +1,107 @@
...
-        params = {
-            "APICAST_NAME": self.deployment,
-            "AMP_APICAST_IMAGE": self.image,
-            "DEPLOYMENT_ENVIRONMENT": "production",
-            "CONFIGURATION_LOADER": "boot",
-            "CONFIGURATION_CACHE": 300,
-            "LOG_LEVEL": "info",
-            "CONFIGURATION_URL_SECRET": self.configuration_url_secret_name,
-        }
+class TemplateApicast(ExternalApicast):
```
mganisin, line 14: now I am confused. `ExternalApicast` is presented as already deployed apicast, this not the case of `TemplateApicast`.
??

testsuite/tests/conftest.py
```diff
@@ -377,12 +377,10 @@ def _custom_user(params, autoclean=True):
 
 
 @pytest.fixture(scope="session")
-def staging_gateway(request, testconfig, configuration):
+def default_staging_gateway(request, testconfig, configuration, openshift):
```
mganisin, line 380: I didn't get concept of `default_staging_gateway`, don't understand, considering it pointless (explanation will fix that)
Also recurring clones of `def staging_gateway` aren't clear to me

testsuite/gateways/apicast/external.py
```diff
@@ -0,0 +1,119 @@
...
+from testsuite.capabilities import Capability
+from testsuite.gateways.apicast import AbstractApicast
+from testsuite.openshift.client import OpenShiftClient
+from testsuite.openshift.env import Environ
+from testsuite.openshift.objects import Routes
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ExternalApicast(AbstractApicast):
```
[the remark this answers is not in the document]
pehala: I like the `IndependentApicast` option

testsuite/gateways/apicast/operator.py
```diff
@@ -1,69 +1,43 @@
...
-        super().on_service_delete(service)
-        del self.openshift.routes[self._route_name(service.entity_id)]
-
     def reload(self):
         self.openshift.do_action("delete", ["pod", "--force",
-                                            "--grace-period=0", "-l", f"deployment={self.name}"])
+                                            "--grace-period=0", "-l", f"deployment={self.deployment}"])
         # pylint: disable=protected-access
-        self.openshift.wait_for_ready(self.name)
+        self.openshift.wait_for_ready(self.deployment)
```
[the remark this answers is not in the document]
pehala: In most gateway `self.deployment` is just an alias to the `self.name`, the reason being that they do not always have to be the same. This is exactly this case as `self.name` refers to the name of the CRD (and the constructor accepts only `self.name` everywhere and deployment is derived), which differs from the actual name of the deployment the CRD generates.

I agree with calling it `self.deployment_name` instead.

testsuite/tests/conftest.py
pehala: The concept was that in many tests we duplicated the code from main conftest when we needed default gateway, since the code wasn´t trivial in previous iterations of this PR, I wanted to simplify it and have less copy-paste there, however through many iterations, the complicated initialization that once was in `staging_gateway` is now gone and so this is now pointless and I will remove it.

testsuite/gateways/apicast/template.py
pehala: `ExternalApicast` expects the Apicast to be already deployed, however, the `TemplateApicast` and `OperatorApicast` extend this behaviour and add the implementation for creation/deletion.

mganisin: then we are back at the beginning, it is neither `ExternalApicast` not `IndependentApicast`, it is `SelfmanagedApicast`

testsuite/gateways/apicast/operator.py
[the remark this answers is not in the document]
mganisin: deployment name from was "typo", of course I wanted to write deployment config. Obviously I was not suggesting deployment_name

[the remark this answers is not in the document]
pehala: The operator uses Deployment instead of DeploymentConfig, so I would stick with `deployment`

mganisin (approved):

JaurbanRH (approved):

PR #16, opened 2021-11-10
Add option for Nav Steps to use Classes as destinations

jsmolar:


testsuite/ui/navigation/step_processor.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    This class is responsible for processing step methods on current view with respect to
+    """
+
+    def __init__(self, current_view):
+        """
+
+        """
+        self.current_view = current_view
+        View_step = namedtuple('Step', 'method destination')
```
pehala, line 64: This seems to be a data class with only one parameter, why you don't pass the parameter instead of creating additional objects?

testsuite/ui/navigation/step_processor.py
```diff
@@ -0,0 +1,84 @@
...
+        :param possible_steps: list of tuples [(method name, method)]. Describes all View methods
+            decorated with @step
+        :param destination: View class
+        :param kwargs: parameters that are passed to particular steps. E.g. account_id={id}
+        :return: Bool value representing if step was preformed or not
+        """
+        rules = [view_class, view_class_string, href]
+        logging.debug(f"Navigator: finding step from {self.current_view} to {desired_dest}")
+        for r in rules:
+            if [step for step in self.view_steps if r(step, desired_dest, **kwargs)]:
```
pehala, line 82: I dont really understand what this if does. `any(r(step, desired_dest, **kwargs) for step in self.view_steps)`

testsuite/ui/navigation/step_processor.py
```diff
@@ -0,0 +1,84 @@
+import inspect
```
pehala, line 1: I don't think we need additional file for this class as it is only used only in the Navigator and is not general enough to be used on its own

pehala (approved):
LGTM! However, I did not run all the UI tests to make sure it works.

testsuite/ui/navigation/exception.py
```diff
@@ -0,0 +1,30 @@
...
+    """Exception when navigations can't be found"""
+
+    def __init__(self, current, dest, possibilities):
+        super().__init__(self)
+        self.current = current
+        self.dest = dest
+        self.possibilities = possibilities
+
+    def __str__(self):
+        return (
```
JaurbanRH, line 14: Could you use f-string instead of format?

testsuite/ui/navigation/exception.py
```diff
@@ -0,0 +1,30 @@
...
+class NavigationStepException(Exception):
+    """Exception when navigation fails on `@step` method"""
+
+    def __init__(self, step):
+        super().__init__(self)
+        self.step = step
+
+    def __str__(self):
+        return (
+            "FAILED: Step method invocation: [{}]"
```
JaurbanRH, line 29: Same here.

PR #17, opened 2021-11-10
Master tenant

dlaso99:
requires [Add billing endpoints for tenant in master api #106](https://github.com/3scale-qe/3scale-api-python/pull/106) to be merged first.

Since there is no way how to retrieve a tenant (3scale API does not support it right now), the `test_create_tenant` has to work only with `account` and `test_edit_tenant` with `test_delete_tenant` are using `custom_tenant` which returns `Tenant` resource. Since `tenants` cant be retrieved, the `account` needs to be retrieved from the tenant.

### DIFF
--- a/testsuite/tests/ui/conftest.py
+++ b/testsuite/tests/ui/conftest.py
@@ -22,6 +22,9 @@ from testsuite.ui.views.admin.login import LoginView
 from testsuite.ui.views.admin.product.application import ApplicationPlanNewView
 from testsuite.ui.views.admin.product.product import ProductNewView
 from testsuite.ui.views.devel.login import LoginDevelView
+from testsuite.ui.views.master.audience.tenant import TenantNewView
+from testsuite.ui.views.master.foundation import BaseMasterView
+from testsuite.ui.views.master.login import MasterLoginView
 from testsuite.ui.webdriver import SeleniumDriver
 from testsuite.utils import blame
 
@@ -87,6 +90,25 @@ def login(custom_admin_login):
     return custom_admin_login()
 
 
+@pytest.fixture
+def master_login(browser, navigator):
+    """
+    Login to the Master portal with default admin credentials
+    :param browser: Browser instance
+    :param navigator: Navigator Instance
+    :return: Login with default credentials
+    """
+    url = settings["threescale"]["master"]["url"]
+    name = settings["threescale"]["master"]["username"]
+    password = settings["threescale"]["master"]["password"]
+    browser.url = url
+
+    page = navigator.open(MasterLoginView)
+
+    if page.is_displayed:
+        page.do_login(name, password)
+
+
 @pytest.fixture(scope="module")
 def custom_devel_login(browser, sessions, navigator, provider_account, account_password):
     """
@@ -137,10 +159,45 @@ def navigator(browser):
         :return: Navigator instance
     """
     base_views = [
-        BaseAdminView
+        BaseAdminView,
+        BaseMasterView
     ]
     navigator = Navigator(browser, base_views)
     return navigator
+
+
+# pylint: disable=unused-argument, too-many-arguments
+@pytest.fixture(scope="module")
+def custom_ui_tenant(navigator, threescale, testconfig, request, master_threescale):
+    """Parametrized custom Tenant created via UI"""
+
+    def _custom_ui_tenant(username: str = "",
+                          email: str = "",
+                          password: str = "",
+                          organisation: str = "",
+                          autoclean=True):
+
+        tenant = navigator.navigate(TenantNewView)
+
+        tenant.create(username=username, email=email+"@anything.invalid", password=password, organization=organisation)
+
+        account = master_threescale.accounts.read_by_name(organisation)
+        tenant = master_threescale.tenants.read(account.entity_id)
+        tenant.wait_tenant_ready()
+
+        if autoclean and not testconfig["skip_cleanup"]:
+            request.addfinalizer(tenant.delete)
+
+        return tenant
+
+    return _custom_ui_tenant
+
+
+@pytest.fixture(scope="module")
+def ui_tenant(custom_ui_tenant, request):
+    """Preconfigured tenant existing over whole testing session"""
+    name = blame(request, "ui-tenant")
+    return custom_ui_tenant(username=name, email=name, password="12345678", organisation=name)
 
 
 # pylint: disable=unused-argument, too-many-arguments
--- a/testsuite/tests/ui/master/test_master_dashboard.py
+++ b/testsuite/tests/ui/master/test_master_dashboard.py
@@ -1,0 +1,18 @@
+"""
+    rewrite of admin dashboard test to master (testing the master dashboard)
+
+    Tests in this file are testing the functionality of the master dashboard.
+"""
+from testsuite.ui.views.master.foundation import MasterDashboardView
+
+
+# pylint: disable=unused-argument
+def test_dashboard_is_loaded_correctly(master_login, navigator):
+    """
+    Test:
+        - Navigates to Dashboard
+        - Checks whether everything is on the dashboard that is supposed to be there
+    """
+    dashboard = navigator.navigate(MasterDashboardView)
+
+    assert dashboard.is_displayed
--- a/testsuite/tests/ui/master/test_tenant.py
+++ b/testsuite/tests/ui/master/test_tenant.py
@@ -1,0 +1,119 @@
+"""
+    Tests that test the tenant modification such as creation, deletion, edit or resume.
+"""
+import pytest
+
+from testsuite.ui.views.admin.login import LoginView
+from testsuite.ui.views.common.foundation import NotFoundView
+from testsuite.utils import blame
+from testsuite.ui.utils import assert_displayed_in_new_tab
+from testsuite.ui.views.master.audience.tenant import TenantDetailView, TenantEditView
+from testsuite.ui.views.devel import LandingView
+
+
+@pytest.fixture(scope="module")
+def tenant_name(request):
+    """ creates a simple blame string """
+    return blame(request, "org-name")
+
+
+@pytest.fixture(scope="module")
+def ui_tenant(tenant_name, custom_ui_tenant):
+    """ overrides the ui_tenant fixture and creates and returns new tenant """
+    return custom_ui_tenant(username=tenant_name, email=tenant_name, password="12345678", organisation=tenant_name)
+
+
+# pylint: disable=too-many-arguments, unused-argument
+def test_create_tenant(master_login, ui_tenant, tenant_name, navigator, browser, master_threescale):
+    """
+    Test:
+        - Creates tenant via UI
+        - Checks whether it exists
+    """
+    account_id = ui_tenant.entity['signup']['account']['id']
+    account = master_threescale.accounts.read(account_id)
+
+    assert account.entity_name == tenant_name
+
+    detail_view = navigator.navigate(TenantDetailView, account=account)
+
+    assert detail_view.public_domain.text == account.entity['domain']
+    assert detail_view.admin_domain.text == account.entity['admin_domain']
+
+    assert_displayed_in_new_tab(browser, detail_view.open_public_domain, LandingView)
+    assert_displayed_in_new_tab(browser, detail_view.open_admin_domain, LoginView)
+
+
+# pylint: disable=unused-argument
+def test_edit_tenant(master_login, navigator, custom_tenant, master_threescale):
+    """
+    Test:
+        - Create tenant via API
+        - Edit tenant via UI
+        - check whether it was edited
+    """
+    tenant = custom_tenant()
+    account_id = tenant.entity['signup']['account']['id']
+    account = master_threescale.accounts.read(account_id)
+
+    old_name = account.entity_name
+    edit = navigator.navigate(TenantEditView, account=account)
+
+    edit.update(org_name="updated_name")
+    account = master_threescale.accounts.read(account_id)
+
+    assert account.entity_name != old_name
+    assert account.entity_name == "updated_name"
+
+
+# pylint: disable=unused-argument
+def test_delete_tenant(master_login, navigator, master_threescale, custom_tenant, browser):
+    """
+    Test:
+        - Create tenant via API without auto-clean
+        - Delete tenant via UI
+        - Assert that deleted tenant is deleted
+    """
+
+    tenant = custom_tenant(autoclean=False)
+    tenant.wait_tenant_ready()
+    account_id = tenant.entity['signup']['account']['id']
+    account = master_threescale.accounts.read(account_id)
+
+    edit = navigator.navigate(TenantEditView, account=account)
+    edit.delete()
+
+    detail_view = navigator.navigate(TenantDetailView, account=tenant)
+
+    account_deleted = master_threescale.accounts.read_by_name(account.entity_name)
+    assert account_deleted.entity['state'] == 'scheduled_for_deletion'
+
+    assert_displayed_in_new_tab(browser, detail_view.open_public_domain, NotFoundView)
+    assert_displayed_in_new_tab(browser, detail_view.open_admin_domain, NotFoundView)
+
+
+# pylint: disable=unused-argument
+def test_resume_tenant(master_login, navigator, master_threescale, custom_tenant, browser):
+    """
+    Test:
+        - Create and Delete tenant via API
+        - Resume the deleted tenant.
+        - Assert that the resume was successful
+    """
+
+    tenant = custom_tenant()
+    tenant.wait_tenant_ready()
+    account_id = tenant.entity['signup']['account']['id']
+    account = master_threescale.accounts.read(account_id)
+    tenant.delete()
+
+    detail_view = navigator.navigate(TenantDetailView, account=tenant)
+
+    # resume tenant from deletion
+    detail_view.resume()
+
+    account_deleted = master_threescale.accounts.read_by_name(account.entity_name)
+    assert account_deleted.entity['state'] != 'scheduled_for_deletion'
+
+    assert_displayed_in_new_tab(browser, detail_view.open_public_domain, LandingView)
+    assert_displayed_in_new_tab(browser, detail_view.open_admin_domain, LoginView)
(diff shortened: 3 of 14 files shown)


dlaso99:
Needs to be merged first -> https://github.com/3scale-qe/3scale-api-python/pull/110

testsuite/tests/ui/master/test_master_dashboard.py
```diff
@@ -0,0 +1,18 @@
+"""
+    rewrite of admin dashboard test to master
```
mdujava, line 2: I would put here *what* is this tests about, not *how* they were created

testsuite/tests/ui/test_login_page.py
```diff
@@ -58,8 +58,8 @@ def test_log_with_random_username_password(refreshed_browser):
     Test Login button is enabled after filling form and expect incorrect credentials
     """
     login_view = LoginView(refreshed_browser)
-    login_view.username_field.fill("username")
-    login_view.password_field.fill("password")
-    assert login_view.submit.is_enabled
-    login_view.submit.click()
+    login_view.login_widget.username_field.fill("username")
```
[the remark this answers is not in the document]
dlaso99: well, the point of `login_widget` is to keep the Login form portion in both Admin and Master Login. Therefore, the split into the `LoginWidget`, which contains the form was necessery.

testsuite/ui/views/master/login.py
```diff
@@ -0,0 +1,32 @@
+""" Representation of Login specific MASTER views"""
+from widgetastic.widget import View, Text
+
+from testsuite.ui.navigation import Navigable
+from testsuite.ui.views.admin.login import LoginWidget
+
+
+class MasterLoginView(View, Navigable):
```
pehala, line 8: It seems to me this Login view shares more with admin `LoginView` than just login widget, wouldn't it make more sense to have common BaseLoginView and both admin/master extending it? @jsmolar @dhlavac WDYT?

jsmolar: We agreed that we don't want to leverage inheritance between admin and master pages.

jsmolar: @dlaso99 remove `Navigable`. Login view should not be navigated to. It is wrongly used in the admin login view as well.

testsuite/ui/views/master/foundation.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class BaseMasterView(View, Navigable):
+    """
+        Basic representation of logged in master portal page.
+        All master portal page should inherits from this class.
+    """
+    path_pattern = ''
+    explorer_menu = Text("//div[@id='api_selector']//a[@title='Context Selector']/span")
+    threescale_menu_logo = GenericLocatorWidget('//*[@id="user_widget"]/a/div')
```
jsmolar, line 20: I suggest that we should wrap this to widget

testsuite/ui/views/devel/__init__.py
```diff
@@ -62,6 +63,22 @@ def post_navigate(self, **kwargs):
         if self.close_csm.is_displayed:
             self.close_csm.click()
 
+    @property
+    def tenant_ready(self) -> bool:
```
jsmolar, line 67: Lets wait for tenant ready defined in API python

mganisin:
Looking at commit structure, this also deserve an update, hint: commits should be feature complete (one PR can contain more features though), certainly commit of message "Rewrite of the previous repo branch" should have zero chance to get merged

testsuite/ui/views/common/foundation.py
```diff
@@ -0,0 +1,17 @@
+"""
```
jsmolar, line 1: I don't like the name: "common" for the package.

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,107 @@
+"""
+    Testing ground for testing master on tenant creation / edit / deletion ...
```
jsmolar, line 2: I don't understand this doc

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+from testsuite.ui.views.admin.login import LoginView
+from testsuite.ui.views.common.foundation import NotFoundView
+from testsuite.utils import blame
+from testsuite.ui.views.master.audience.tenant import TenantDetailView, TenantEditView
+from testsuite.ui.views.devel import LandingView
+
+
+# pylint: disable=invalid-name
+def check_is_displayed_in_new_tab(browser, new_tab_opener, View, wait=False):
```
jsmolar, line 12: This is not "check" but "assert". 
Why are you using `View` and not `view -> View`?
Do you really need param `wait` here?

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,107 @@
...
+        wait for the website to be loaded by refreshing it. (Example -> LandingView)
+    """
+    with browser.new_tab(new_tab_opener):
+        view = View(browser)
+        if wait:
+            view.wait_page_ready()
+        assert view.is_displayed
+
+
+def test_create_tenant(custom_ui_tenant, request, navigator, browser, master_threescale):
```
jsmolar, line 25: Can you use fixture ui_tentant?

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,107 @@
...
+# pylint: disable=unused-argument
+def test_delete_tenant(master_login, navigator, master_threescale, custom_tenant, browser):
+    """
+    Test:
+        - Create tenant via API without auto-clean
+        - Delete tenant via UI
+        - Assert that deleted tenant is deleted
+    """
+
+    tenant = custom_tenant(autoclean=False)
```
jsmolar, line 78: Why disabling autoclean?

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,107 @@
...
+    assert account_deleted.entity['state'] != 'scheduled_for_deletion'
+
+    check_is_displayed_in_new_tab(browser, detail_view.open_public_domain, LandingView)
+    check_is_displayed_in_new_tab(browser, detail_view.open_admin_domain, LoginView)
+
+    edit = navigator.navigate(TenantEditView, account=tenant)
+    edit.delete()
+
+    account_deleted = master_threescale.accounts.read_by_name(account.entity_name)
+    assert account_deleted.entity['state'] == 'scheduled_for_deletion'
```
jsmolar, line 107: Why are you using two same asserts in one test? I think this should be two separate tests: delete and "resume tenant from deletion". But I would start with just one test: delete tenant. The problem with the resume tenant is that you would need to test if it works correctly when resumed.

jsmolar (changes requested):

dlaso99:
The reasoning behind `autoclean=False` in the `test_delete_tenant` is that if the auto-clean is kept on, the deletion will result in an error. Since the test is deleting the tenant, it seems reasonable to set the `autoclean` to `false`.   @jsmolar this is the answer for your question here -> https://github.com/3scale-qe/3scale-tests/pull/17#discussion_r752290668 and also by doing this, the `teardown` problem mentioned in a meeting was resolved

testsuite/tests/ui/master/test_tenant.py
dlaso99: Please see https://github.com/3scale-qe/3scale-tests/pull/17#issuecomment-1047680286

testsuite/tests/ui/master/test_tenant.py
dlaso99: `wait` is not needed and the `View` was renamed after the consultation about the naming conventions with @pehala

testsuite/tests/ui/master/test_tenant.py
dlaso99: an edit to this doc was made, to make it a bit more clear.

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,117 @@
...
+        - Edit tenant via UI
+        - check whether it was edited
+    """
+    tenant = custom_tenant()
+    account_id = tenant.entity['signup']['account']['id']
+    account = master_threescale.accounts.read(account_id)
+
+    edit = navigator.navigate(TenantEditView, account=account)
+
+    edit.update(org_name="updated_name")
```
jsmolar, line 61: assertion, that value was changed is necessary.

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,117 @@
...
+
+    edit = navigator.navigate(TenantEditView, account=account)
+    edit.delete()
+
+    detail_view = navigator.navigate(TenantDetailView, account=tenant)
+
+    account_deleted = master_threescale.accounts.read_by_name(account.entity_name)
+    assert account_deleted.entity['state'] == 'scheduled_for_deletion'
+
+    check_is_displayed_in_new_tab(browser, detail_view.open_public_domain, NotFoundView)
```
jsmolar, line 89: This should be a custom assert method.. What do you think @pehala ?

testsuite/ui/views/common/login.py
```diff
@@ -0,0 +1,31 @@
+"""Login portion which is the same for admin and master"""
+from widgetastic.widget import TextInput, View, Text
+
+from testsuite.ui.widgets.buttons import ThreescaleSubmitButton
+
+
+class LoginWidget(View):
```
jsmolar, line 7: This is View, not Widget! In addition, I suggest that we name it more expressive .. `LoginForm`

testsuite/ui/views/master/audience/tenant.py
```diff
@@ -0,0 +1,137 @@
...
+from widgetastic.widget import TextInput, GenericLocatorWidget, Text
+
+from testsuite.ui.navigation import step
+from testsuite.ui.views.master.audience import BaseMasterAudienceView
+from testsuite.ui.widgets import AudienceTable
+from testsuite.ui.widgets.buttons import ThreescaleDeleteButton, \
+    ThreescaleEditButton, ThreescaleSubmitButton, ThreescaleSearchButton
+
+
+class TenantView(BaseMasterAudienceView):
```
jsmolar, line 11: `TenantsView`

testsuite/ui/views/master/audience/tenant.py
```diff
@@ -0,0 +1,137 @@
...
+        return BaseMasterAudienceView.is_displayed.fget(self) and self.new_account.is_displayed and \
+               self.table.is_displayed and self.path_pattern in self.browser.url
+
+
+class TenantDetailView(BaseMasterAudienceView):
+    """View representation of Tenant detail page"""
+    path_pattern = '/buyers/accounts/{account_id}'
+    edit_button = ThreescaleEditButton()
+    applications_button = Text("//*[contains(@title,'applications')]")
+    public_domain = Text('//*[@id="twoCol"]/div[1]/div/table/tbody/tr[2]/td/a')
```
jsmolar, line 49: I don't like this xpath

testsuite/ui/views/master/audience/tenant.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __init__(self, parent, account):
+        super().__init__(parent, account_id=account.entity_id)
+
+    @step("TenantEditView")
+    def edit(self):
+        """Edit account"""
+        self.edit_button.click()
+
+    def suspend_or_resume(self):
```
jsmolar, line 61: We need a method that would just suspend or just resume the tenant. This is not deterministic.

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+from testsuite.ui.views.admin.login import LoginView
+from testsuite.ui.views.common.foundation import NotFoundView
+from testsuite.utils import blame
+from testsuite.ui.views.master.audience.tenant import TenantDetailView, TenantEditView
+from testsuite.ui.views.devel import LandingView
+
+
+# pylint: disable=invalid-name
+def check_is_displayed_in_new_tab(browser, new_tab_opener, view_class):
```
jsmolar, line 12: Did you ask Peto about typing?

pehala (commented):
Please edit the commit history as well, a few commits seem to be redundant

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,117 @@
...
```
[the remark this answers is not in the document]
dlaso99: Utils file created in `testsuite/ui`

testsuite/tests/ui/master/test_tenant.py
dlaso99: changed the name of the method so it fits the convensions

testsuite/tests/ui/master/test_master_dashboard.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+# pylint: disable=unused-argument
+def test_dashboard_is_loaded_correctly(master_login, navigator):
+    """
+    Test:
+        - Navigates to Dashboard
+        - Checks whether everything is on the dashboard that is supposed to be there
+
+        if this fails, there is higher chance that there is something wrong in the core, rather than here.
```
jsmolar, line 16: Why we need this line?

testsuite/tests/ui/conftest.py
```diff
@@ -143,6 +180,50 @@ def navigator(browser):
...
+        return tenant
+
+    return _custom_ui_tenant
+
+
+@pytest.fixture(scope="module")
+def ui_tenant(custom_ui_tenant, request):
+    """Preconfigured tenant existing over whole testing session"""
+    name = blame(request, "ui_tenant")
+    return custom_ui_tenant(name, name)
```
jsmolar, line 224: I think, this will not work

dlaso99: U were right, it did not work for 2 reasons:

1. you cant have `_` in the forms
2. I was not providing the 4th (the most important) argument :D 

FIxed now + tested

testsuite/ui/views/master/audience/tenant.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class TenantDetailView(BaseMasterAudienceView):
+    """View representation of Tenant detail page"""
+    path_pattern = '/buyers/accounts/{account_id}'
+    edit_button = ThreescaleEditButton()
+    applications_button = Text("//*[contains(@title,'applications')]")
+    public_domain = Text(".//th[contains(text(),'Public domain')]/parent::*/td/a")
+    admin_domain = Text(".//th[contains(text(),'Admin domain')]/parent::*/td/a")
+    suspend_button = Text('//*[@id="twoCol"]/div[1]/div/table/tbody/tr[6]/td/a')
```
dhlavac, line 51: I would recommend separating suspend and resume buttons and using the PF4 button widget for it. E.g. `Button("Resume", classes=[Button.LINK])`

dhlavac (changes requested):

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,118 @@
...
+    account = master_threescale.accounts.read(account_id)
+
+    old_name = account.entity_name
+    edit = navigator.navigate(TenantEditView, account=account)
+
+    edit.update(org_name="updated_name")
+    account = master_threescale.accounts.read(account_id)
+
+    assert account.entity_name == "updated_name"
+    assert account.entity_name != old_name
```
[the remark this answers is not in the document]
dlaso99: I agree, the second assert was added after the @jsmolar pointed out that I am not actually checking if there was a change and that the first one is just checking whether the name is set to some constant. I think that this additional assert just adds a secondary check in case that the accounts name is actually the same as we are changing it to (Undefined behavior?)

If you think that one of them should be removed, which one do you recommend ?

testsuite/tests/ui/master/test_tenant.py
```diff
@@ -0,0 +1,118 @@
...
+from testsuite.utils import blame
+from testsuite.ui.utils import assert_displayed_in_new_tab
+from testsuite.ui.views.master.audience.tenant import TenantDetailView, TenantEditView
+from testsuite.ui.views.devel import LandingView
+
+
+@pytest.fixture(scope="module")
+def custom_blame(request):
+    """ creates a simple blame string """
+    return blame(request, "org-name")
```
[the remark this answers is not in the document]
jsmolar: You are not using `blame` here. You want to generate a random name that is used in your tests and you can verify if specific resources contain this "blamed name". The name of this fixture is vrong.

testsuite/tests/ui/master/test_tenant.py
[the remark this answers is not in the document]
jsmolar: As Laso wrote. The second assertion is to verify if the name was really changed.

testsuite/tests/ui/master/test_tenant.py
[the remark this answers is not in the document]
dlaso99: Renaming the `custom_blame` -> `tenant_name`

testsuite/tests/ui/master/test_tenant.py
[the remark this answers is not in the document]
dhlavac: I am not against, double-checking being better than nothing but I cannot imagine any possible scenario where `account.entity_name == "updated_name"`  check will pass but second check `assert account.entity_name != old_name` will fail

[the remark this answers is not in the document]
dlaso99: before - `updated_name`
after - `updated_name`

Altho this is imho the "Undefined behavior" :/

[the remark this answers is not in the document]
dhlavac: You are always creating a new tenant that has a unique generated name.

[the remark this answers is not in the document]
pehala: How about just switching the order of the asserts? First if it changed and second if it changes to the correct value?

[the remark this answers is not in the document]
dlaso99: I like the @pehala 's idea ... what do you think @dhlavac ?

dhlavac (approved):

jsmolar (approved):

Issue #10, opened 2021-11-05
Mitigate threescale_api.errors.ApiClientError: Response(409 Conflict): b''


mganisin:

mganisin:
Fresh deployment of 3scale still requires some time to initialize returns some failures menawihle. A wating action is needed to ensure all is up and running

mdujava:
Same can happen on https://github.com/3scale-qe/3scale-tests/blob/6588732634d4c15b02d5b44f9bebf8d290dc1c57/testsuite/tests/custom_tenant/test_custom_tenant.py#L41

PR #31, opened 2021-11-12
added backoff to mitigate ApiClientError(409)

palmieric:
This PR fix the [ISSUE#10](https://github.com/3scale-qe/3scale-tests/issues/10)

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -8,8 +8,7 @@ import time
 import importlib_resources as resources
 import backoff
 import pytest
-import threescale_api.errors
-from threescale_api import client
+from threescale_api import client, errors
 from weakget import weakget
 
 # to actually initialize all the providers
@@ -283,6 +282,14 @@ def account_password():
     return '123456'
 
 
+@backoff.on_exception(backoff.fibo, errors.ApiClientError, 8, jitter=None)
+def call_account(custom_account, params):
+    """A wrap around custom_account function to backoff the request
+    on Exception 'threescale_api.errors.ApiClientError: Response(409 Conflict)'
+    """
+    return custom_account(params=params)
+
+
 @pytest.fixture(scope="session")
 def account(custom_account, request, testconfig, account_password):
     "Preconfigured account existing over whole testing session"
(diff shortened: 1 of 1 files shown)


testsuite/tests/conftest.py
```diff
@@ -797,7 +804,7 @@ def _custom_tenant(name="t", autoclean=True, wait=True):
 
             unprivileged_client = client.ThreeScaleClient(admin_base_url, access_token, ssl_verify=False)
 
-            @backoff.on_exception(backoff.fibo, threescale_api.errors.ApiClientError, max_tries=8, jitter=None)
+            @backoff.on_exception(backoff.fibo, errors.ApiClientError, max_tries=8, jitter=None)
```
mganisin, line 807: This looks like unrelated and unnecessary change

testsuite/tests/conftest.py
```diff
@@ -283,6 +282,14 @@ def account_password():
     return '123456'
 
 
+@backoff.on_exception(backoff.fibo, errors.ApiClientError, 8, jitter=None)
+def call_account(custom_account, params):
```
mganisin, line 286: a) maybe I would prefer this to be an inner function inside `def account` to hide it as "implementation detail"
b) should this be rather in `def custom_account` with new parameter `wait` this time by default set to `True`?
c) (just a comment) this catches `ApiClientError` to broadly, initially I was thinking just about checking the presence of 409 status, however as it is it may also solve the other issues raised during custom tenant, so keep it as it is. On the other hand all this will slowdown significantly the execution of all tests if something will be wrong with 3scale (and this will raise an error every time due to that). This was the reason I wanted it as a separate function to be called before the tests (now recalled the idea). Keep it as it is for now, but if we will hit that problem (and in such case it will be clearly visible), this will have to be reimplemented

mganisin (changes requested):

testsuite/tests/conftest.py
palmieric: a) ack
b) ack
c) I slightly disagree with this, the test execution slows down only if there are issues in 3scale API availability, but if this is the case the execution is slow or failed in any case.

testsuite/tests/conftest.py
palmieric: I will submit this in a different PR

mganisin (approved):

mdujava (approved):

PR #34, opened 2021-11-12
Add sleep into auth0 setup due to slow auth0 API

JaurbanRH:
Sometimes Auth0 API is slow and didn't set callback urls in time hence we need to add sleep.


testsuite/tests/ui/devel/auth/test_login_auth0.py
```diff
@@ -38,6 +40,8 @@ def auth0_setup(custom_admin_login, navigator, testconfig, set_callback_urls, au
 
     sso = navigator.navigate(Auth0IntegrationDetailView, integration=auth0_integration)
     set_callback_urls(testconfig["auth0"]["client"], sso.callback_urls())
+    sleep(3)  # Sometimes Auth0 API is slow and didn't set callback urls in time
```
mganisin, line 43: looks like sleep should be in like set_callback_urls

mganisin (changes requested):

pehala:
Can´t we use backoff until it is there?

testsuite/tests/ui/conftest.py
```diff
@@ -387,8 +388,10 @@ def set_callback_urls(auth0_client):
     Set callback urls for Auth0 application
     """
 
+    @backoff.on_predicate(backoff.constant, lambda x: x['callbacks'] == [], 10)
```
mganisin, line 391: sleep(3) was replaced by constant backoff of count 10? doesn't look equal (plus missing jitter=None ...)

mganisin (approved):

mdujava (approved):

PR #35, opened 2021-11-12
Draft:Removed xfail marks on fixed issues

dhlavac:
SSIA


mganisin (approved):

testsuite/tests/ui/devel/auth/test_login_rhsso.py
```diff
@@ -61,6 +60,4 @@ def test_devel_login_rhsso(custom_devel_rhsso_login, navigator, testconfig):
     signup_view = SignUpView(navigator.browser)
```
JaurbanRH, line 60: Next 2 lines should be removed as well since you are not working with SignupView in this test at all.

pehala (approved):

PR #51, opened 2021-11-26
Custom tenat test 

mdujava:
There is backoff in a test so we don't need to wait in fixture.
Updating the reason for mark.flaky.

### DIFF
--- a/testsuite/tests/custom_tenant/test_custom_tenant.py
+++ b/testsuite/tests/custom_tenant/test_custom_tenant.py
@@ -2,41 +2,23 @@
 rewritten the /spec/functional_specs/custom_tenant_spec.rb
 Creates a new tenant and tests if the tenant works
 """
-import backoff
 import pytest
-from threescale_api import client
 
 
 @pytest.fixture(scope="session")
-def threescale(custom_tenant, testconfig):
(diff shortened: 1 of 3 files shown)


testsuite/tests/custom_tenant/test_custom_tenant.py
```diff
@@ -37,6 +37,9 @@ def get_custom_client(url, access_token, ssl_verify):
 
 # FIXME: threescale_api.errors.ApiClientError: Response(422): b'{"errors":{"system_name":["must be shorter."]}}
 # Name of the tenant is too long
+# ^ no longer occuring
+# FIXME: threescale_api.errors.ApiClientError: Response(404 Not Found): b'{"status":"Not found"}'
```
pehala, line 41: Would it be possible to override `custom_application` just for this test to add backoff there too?

mdujava: I would discuss it in https://github.com/3scale-qe/3scale-tests/pull/57, what do you think?

pehala: Well, not sure if this should be testsuite wise, I would say this is only for this test

PR #55, opened 2021-11-29
Implement environmental varables in OperatorApicast through CRD

pehala:
* It is possible to set/get environmental variables in `OperatorApicast` through `environ()` method.
   * CRD is used instead of directly changing the deployment
   * delete is not yet supported, since it is not used at all
*  Not all environmental variables are implemented yet
   * This will be done in subsequent PRs
* Extracted `Properties` interface from `Environ` and made `environ()` method return `Properties`
* Add `OperatorEnviron` which handles transforming from env vars to CRD names
* Changed tests to use `SelfManagedApicast` instead of `TemplateApicast` where possible.
   * This means the tests will use `OperatorApicast` on OCP4 and `TemplateApicast` on ocp3  
   * Choose opt-in approached as it will be clearly visible from PRs which tests are now supported on OCP4
 
 If the number of files this PR changes is too much, it is possible to split this in two, where one is the first commit only.

### DIFF
--- a/testsuite/openshift/crd/apicast.py
+++ b/testsuite/openshift/crd/apicast.py
@@ -62,3 +62,6 @@ class APIcast(APIObject):
 
     def __setitem__(self, key, value):
         self.model.spec[key] = value
+
+    def __delitem__(self, key):
+        del self.model.spec[key]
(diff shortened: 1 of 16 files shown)


mganisin (dismissed):

testsuite/openshift/crd/apicast.py
```diff
@@ -56,9 +56,19 @@ def commit(self):
         self.create()
         return self.refresh()
 
+    def apply(self, cmd_args=None):
+        """
+        Workaround until https://github.com/openshift/openshift-client-python/pull/102 is merged
```
mdujava, line 61: It was merged.

mdujava (changes requested):

mdujava (approved):

mganisin (approved):

PR #57, opened 2021-11-30
Revert backoff on custom_account fixture

mganisin:
The backoff was supposed to mitigate 409 error returned right after the
deployment or custom tenant creation.

This didn't fix it as the API returned 422 on repeated calls and made
the situation even worse under some circumstances. The only efect was
that the faliure was significantly delayed and test execution prolonged.


palmieric:
I tested this against the custom_tenant test around 200 times with a failure rate of 20% with the error code 422.
How often are you seeing this?

PR #77, opened 2021-12-07
Improve custom policy tests

pehala:
* Splits the big custom policy template which builds custom policy image twice (once from Github, once by copying the necessary content) into two smaller templates
   * This enables usage of this template in other custom policy tests with only one build and one image stream instead of two
   * The modular_apicast test now uses only one image stream with two images in it.
* Renames most of the variables inside the template so they are more verbose a hopefully easier to understand.


mganisin (dismissed):
This one is GREAT, exactly what I wanted to see :-)

pehala:
* Removed now unused on_failed template

testsuite/tests/apicast/parameters/test_modular_apicast.py
```diff
@@ -45,34 +36,44 @@ def build_images(openshift, request, image_template,
...
+                   "TARGET_TAG": "latest",
+                   "EXAMPLE_POLICY_IMAGE_STREAM": image_stream_name,
+                   "EXAMPLE_POLICY_TAG": "github"}
 
     def _delete_builds():
-        openshift_client.delete_template(image_template, template_params)
+        openshift_client.delete_template(github_template, github_params)
+        openshift_client.delete_template(copy_template, copy_params)
 
     request.addfinalizer(_delete_builds)
```
palmieric, line 62: I'd just change this to make use of yield instead of finalizer

pehala: I think the finalizer is better in this case as all of the commands below can sometimes fail which would leave many objects undeleted.

mganisin: right, `addfinalizer` seems appropriate here

PR #78, opened 2021-12-07
Introduce fast testing make target

mganisin:
There is new mark 'sandbag' and new target 'speedrun'. Both are supposed
to work together to enable fast test executin by skipping 'troublesome'
tests (slow, complex setup, ...). This will help with testing other
environments.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,7 +2,7 @@
 markers =
 	disruptive: This has side effect with possible impact on other tests
 	performance: Performance tests have unique needs
-	slow: This is time consuming
+	sandbag: Either slow, deployment sensitive or requires complex setup
 	smoke: Build verification test
 	flaky: Random failures with unclear reason
 	required_capabilities(capability1, capability2, ...): List of capabilities that are required for running this test
--- a/testsuite/tests/apicast/policy/websocket/test_websocket_policy_app_id.py
+++ b/testsuite/tests/apicast/policy/websocket/test_websocket_policy_app_id.py
@@ -7,7 +7,10 @@ from threescale_api.resources import Service
 from testsuite import TESTED_VERSION  # noqa # pylint: disable=unused-import
 from testsuite.tests.apicast.policy.websocket.conftest import retry_sucessful, retry_failing
 
-pytestmark = [pytest.mark.skipif("TESTED_VERSION < Version('2.8')")]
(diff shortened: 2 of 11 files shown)


mganisin:
more or less this is just a draft, we probably rather need a `--drop-sandbag` option that will skip `sandbag` as well as some other tests (typically selected capabilities). Winking at @pehala 's direction

mganisin:
Now it is final with `--drop-sandbag` option skipping `sandbag` marks as well as selected `capabilities`

pytest.ini
```diff
@@ -2,7 +2,7 @@
 markers =
 	disruptive: This has side effect with possible impact on other tests
 	performance: Performance tests have unique needs
-	slow: This is time consuming
+	sandbag: Either slow, deployment sensitive or requires complex setup
```
pehala, line 5: I would change the name either to complex or slow, sandbag is not very verbose

testsuite/tests/apicast/policy/websocket/test_websocket_policy_app_id.py
```diff
@@ -7,7 +7,9 @@
 from testsuite import TESTED_VERSION  # noqa # pylint: disable=unused-import
 from testsuite.tests.apicast.policy.websocket.conftest import retry_sucessful, retry_failing
 
-pytestmark = [pytest.mark.skipif("TESTED_VERSION < Version('2.8')")]
+pytestmark = [
+        pytest.mark.sandbag,  # deployment sensitive
```
pehala, line 11: I would like to see why it is deployment sensitive and what it actually needs to be deployed.

testsuite/tests/conftest.py
```diff
@@ -73,6 +77,15 @@ def pytest_runtest_setup(item):
             for capability in mark.args:
                 if capability not in CapabilityRegistry():
                     pytest.skip(f"Skipping test because current environment doesn't have capability {capability}")
+    if item.config.getoption("--drop-sandbag"):
+        sandbag_caps = (Capability.CUSTOM_ENVIRONMENT, Capability.LOGS, Capability.JAEGER)
```
pehala, line 81: I would remove Capability.CUSTOM_ENVIRONMENT as I believe this mark is only applied on some parameters tests and is not inherently slow or requires special deployment

pehala (approved):

palmieric (approved):

PR #84, opened 2021-12-14
Refactor test_email_accounts.py

mganisin:
It is absolutely fine to reject this PR (however great idea to accept it) and the reason is (at least) cleanup in the test itself.
Anyway I still prefer to accept it as it is much easier to understand this version in my opinion (what is very subjective and may be another reason for rejection actually) and should be way much easier to debug (though the reporting/logging is still sub-optimal). Currently the test is flaky, this was the impulse to rewrite it, however I believe I can fix also original test in case of rejection.

It was bit difficult to understand the original code. While this
proposal might be hard to read as well for someone, I believe its flow
is much easier to follow and particular checks should be clearer.

Furthermore fixtures are now module-scoped, this is because of
backoff/flakiness. In this particular case it is not desirable to repeat
setup/cleanup each time test is executed, what ma be a risk in case of
function scoped fixtures.

### DIFF
--- a/testsuite/tests/system/messages/test_email_accounts.py
+++ b/testsuite/tests/system/messages/test_email_accounts.py
@@ -15,7 +15,7 @@ from testsuite import rawobj
 from testsuite.utils import blame
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def application(service, custom_application, custom_app_plan, lifecycle_hooks, request):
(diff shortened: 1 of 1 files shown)


palmieric:
Totally agree, the test is much cleaner now with this refactoring.

palmieric (approved):

pehala (approved):

PR #85, opened 2021-12-14
Change the way amp_release is gathered for custom apicast deployments

mganisin:
Original implementation to gather version from imagestream caused
troubles during upgrades as upgraded deployment had two imagestreams
actually.

### DIFF
--- a/testsuite/tests/apicast/parameters/test_modular_apicast.py
+++ b/testsuite/tests/apicast/parameters/test_modular_apicast.py
@@ -25,7 +25,7 @@ def image_stream_name(request):
 
 
 @pytest.fixture(scope="module")
-def build_images(openshift, request, image_stream_name):
+def build_images(openshift, testconfig, request, image_stream_name):
     """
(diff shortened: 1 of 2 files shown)


pehala:
Just so I don't forget it through Christmas, I believe we may be able to tell the exact version we want by looking at the image triggers in deployment config of apicast-staging. This could theoretically solve our problems but I haven't tested it yet.

mganisin:
> Just so I don't forget it through Christmas, I believe we may be able to tell the exact version we want by looking at the image triggers in deployment config of apicast-staging. This could theoretically solve our problems but I haven't tested it yet.

From where apicast-staging will be deployed? I'll answer myself: from APIManager/3scale-template == No difference from this proposal

pehala:
> From where apicast-staging will be deployed? I'll answer myself: from APIManager/3scale-template == No difference from this proposal

In my too

pehala (approved):

pehala (approved):

JaurbanRH (approved):

PR #87, opened 2021-12-17
Fixed IFrames of Billing SCA pages

jsmolar:


testsuite/ui/browser.py
```diff
@@ -104,3 +107,24 @@ def new_tab(self, trigger, keep_tab=False):
         else:
             self.browser.close_window(new_handle)
             self.browser.switch_to_window(current_handle)
+
+    @backoff.on_exception(backoff.fibo, NoSuchElementException, max_tries=13, jitter=None)
```
JaurbanRH, line 111: I think max_tries=13 is too much. Use e.g. 8 as in rest of testsuite.

jsmolar: 8 was sometimes not enough for the OTP form to load.

dhlavac (approved):

testsuite/ui/views/devel/settings/stripe.py
```diff
@@ -11,7 +11,7 @@
 logger = logging.getLogger(__name__)
 
 
-class BillingAddressForm(View):
+class _BillingAddressForm(View):
```
pehala, line 14: I dont think this name change is needed as it requires module-level pylint ignore.

testsuite/ui/views/devel/settings/stripe.py
```diff
@@ -35,50 +35,57 @@ def add(self, address):
...
+
+class _StripeCCForm(View):
     """Stripe Credit card form"""
-    submit_btn = GenericLocatorWidget("//button[@id='stripe-submit']")
+    submit_btn = GenericLocatorWidget(".//button[@id='stripe-submit']")
+
+    @View.nested
+    # pylint: disable=invalid-name
+    class cc_details(View):
+        """I frame that contains credit card information"""
```
pehala, line 50: IFrame

testsuite/ui/views/devel/settings/stripe.py
```diff
@@ -1,7 +1,7 @@
 """Settings Devel portal View containing credit card details for Stripe payment gateway"""
+# pylint: disable=invalid-name
```
mganisin, line 2: ???

testsuite/ui/views/devel/settings/stripe.py
```diff
@@ -35,50 +35,57 @@ def add(self, address):
...
 
-class StripeCCForm(View):
+
+class _StripeCCForm(View):
     """Stripe Credit card form"""
-    submit_btn = GenericLocatorWidget("//button[@id='stripe-submit']")
+    submit_btn = GenericLocatorWidget(".//button[@id='stripe-submit']")
+
+    @View.nested
+    # pylint: disable=invalid-name
```
mganisin, line 48: is this necessary?

testsuite/ui/views/devel/settings/stripe.py
mganisin: Good question. Why? (though it is valid name in my opinion)

testsuite/ui/views/devel/settings/stripe.py
```diff
@@ -35,50 +35,57 @@ def add(self, address):
...
+    class cc_details(View):
+        """I frame that contains credit card information"""
+        FRAME = ".//form[@id='stripe-form']//iframe"
+        cc_number = TextInput(name="cardnumber")
```
mganisin, line 52: **cardnumber** = TextInput(name="**cardnumber**")

testsuite/ui/views/devel/settings/stripe.py
jsmolar: This one does not require pylint ignore. 
`BillingAddressFroms` is used only by `StripeCCForm` so I assumed that package-private would be correct encapsulation

testsuite/ui/views/devel/settings/stripe.py
jsmolar: What is more preferable? To use module-level pylint ignore or use the same one on 3 different places in a module?

pehala: Definitely 3 in the exact spots they are needed

mganisin (approved):

JaurbanRH (approved):

PR #90, opened 2021-12-22
Repair of auth tests

JaurbanRH:
Add sleep into auth0 setup due to slow auth0 API. Add missing cleanup to auth tests. Change scope of browser fixture for devel auth tests.

### DIFF
--- a/testsuite/tests/ui/auth/conftest.py
+++ b/testsuite/tests/ui/auth/conftest.py
@@ -34,7 +34,8 @@ def ui_sso_integration(custom_admin_login, navigator, threescale, testconfig, re
 
 
 @pytest.fixture(scope="module")
-def auth0_setup(custom_admin_login, testconfig, ui_sso_integration, navigator, set_callback_urls, auth0_user):
+def auth0_setup(custom_admin_login, testconfig, ui_sso_integration, navigator, set_callback_urls, auth0_user,
+                threescale):
     """
     Preparation for Auth0 tests:
         - Crate Auth0 SSO integration
@@ -54,7 +55,12 @@ def auth0_setup(custom_admin_login, testconfig, ui_sso_integration, navigator, s
     assert sso.test_flow_checkbox.is_checked
     sso.publish()
 
-    return urls
+    yield urls
+
+    if not testconfig["skip_cleanup"]:
+        name = auth0_user["email"].split("@")[0]
+        user = threescale.provider_account_users.read_by_name(name)
+        user.delete()
 
 
 @pytest.fixture
@@ -109,7 +115,7 @@ def rhsso_setup(request, custom_admin_login, rhsso_service_info, ui_sso_integrat
 
 
 @pytest.fixture
-def rhsso_login(rhsso_setup, testconfig, custom_rhsso_login):
+def rhsso_login(rhsso_setup, testconfig, custom_rhsso_login, rhsso_service_info):
     """Login into 3scale via RHSSO"""
     test_user = testconfig["rhsso"]["test_user"]
     custom_rhsso_login(test_user["username"], test_user["password"])
@@ -119,6 +125,9 @@ def rhsso_login(rhsso_setup, testconfig, custom_rhsso_login):
 def rhsso_bounce_login(rhsso_setup, navigator, testconfig, browser, rhsso_service_info, threescale):
     """Login into 3scale via RHSSO using /bounce URL"""
     bounce_url = rhsso_setup[0].replace("/callback", "/bounce")
+    username = testconfig["rhsso"]["test_user"]["username"]
+    user_id = rhsso_service_info.realm.admin.get_user_id(username)
+    rhsso_service_info.realm.admin.logout(user_id)
     navigator.open(url=bounce_url)
     provider = RhssoView(browser.root_browser)
     test_user = testconfig["rhsso"]["test_user"]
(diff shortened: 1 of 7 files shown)


testsuite/tests/ui/conftest.py
```diff
@@ -388,10 +388,9 @@ def set_callback_urls(auth0_client):
     Set callback urls for Auth0 application
     """
 
-    @backoff.on_predicate(backoff.fibo, lambda x: not x, 8, jitter=None)
     def _set_callback_urls(client_id, urls: list):
         auth0_client.clients.update(client_id, body={"callbacks": urls})
-        return auth0_client.clients.get(client_id)
+        sleep(15)
```
pehala, line 393: This deserves an explanation

pehala (dismissed):

testsuite/tests/ui/devel/auth/conftest.py
```diff
@@ -3,16 +3,48 @@
...
+    """
+    webdriver = SeleniumDriver(source=settings["fixtures"]["ui"]["browser"]["source"],
+                               driver=settings["fixtures"]["ui"]["browser"]["webdriver"],
+                               ssl_verify=settings["ssl_verify"],
+                               remote_url=settings["fixtures"]["ui"]["browser"]["remote_url"],
+                               binary_path=settings["fixtures"]["ui"]["browser"]["binary_path"])
+    webdriver.get_driver()
+    webdriver.post_init()
+    started_browser = ThreeScaleBrowser(selenium=webdriver.webdriver)
+    request.addfinalizer(webdriver.finalize)
```
mganisin, line 28: in this case addfinalizer might make sense, however likely 3 lines above?

testsuite/tests/ui/auth/conftest.py
```diff
@@ -74,8 +79,7 @@ def auth0_bounce_login(auth0_setup, navigator, browser, auth0_user):
 
 # pylint: disable=too-many-locals
 @pytest.fixture(scope="module")
-def rhsso_setup(request, custom_admin_login, rhsso_service_info, ui_sso_integration, navigator, testconfig,
-                threescale):
+def rhsso_setup(custom_admin_login, rhsso_service_info, ui_sso_integration, navigator, testconfig, threescale):
```
mganisin, line 82: this is almost unrelated change. The goal isn'ŧ to fix "everything", it is about not introducing more "antipaterns". In this particular case, rather than rewrite of finalization it's better to just add missing condition (if I understand the change properly). Keeping commit (and PR) as small as possible containing just relevant changes has highest priority

mganisin (changes requested):

testsuite/tests/ui/auth/conftest.py
```diff
@@ -54,7 +54,12 @@ def auth0_setup(custom_admin_login, testconfig, ui_sso_integration, navigator, s
     assert sso.test_flow_checkbox.is_checked
     sso.publish()
 
-    return urls
+    yield urls
+
+    if not testconfig["skip_cleanup"]:
```
dhlavac, line 59: Currently, test are not cleaning Auth0 auth source only users created by this auth source

dhlavac (dismissed):

dhlavac:
IMHO overall would be good to separate Auth0 and RHSSO tests to separate files/modules etc because currently to test only one of them we need to edit tests manually

pehala:
> IMHO overall would be good to separate Auth0 and RHSSO tests to separate files/modules etc because currently to test only one of them we need to edit tests manually

Separating it would probably lead to useless code duplication.

I believe that if the tests are written correctly (Proper cleaning etc.) it shouldn't matter. If it is needed on regular basis then it should be handled with skipif and if it is needed only once in a while I don't think it is that hard to manually.

testsuite/tests/ui/conftest.py
```diff
@@ -388,10 +389,11 @@ def set_callback_urls(auth0_client):
     Set callback urls for Auth0 application
     """
 
-    @backoff.on_predicate(backoff.fibo, lambda x: not x, 8, jitter=None)
     def _set_callback_urls(client_id, urls: list):
         auth0_client.clients.update(client_id, body={"callbacks": urls})
-        return auth0_client.clients.get(client_id)
+        # Due to slow Auth0 API/UI we need to use sleep here. Backoff is not viable option since the client
+        # updates URLs correctly but UI need to wait some time to correctly retrieve URLs.
```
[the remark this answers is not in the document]
JaurbanRH: After trying to login via Auth0, UI shows error message but when you refresh page after couple of seconds everything works as it should. I think that is easier to use sleep instead of using backoff based on is_displayed on UI element and then refreshing page.

pehala (changes requested):

mganisin (approved):

testsuite/tests/ui/devel/auth/conftest.py
```diff
@@ -3,44 +3,92 @@
...
 from testsuite import settings
+from testsuite.tests.ui import Sessions
+from testsuite.ui.browser import ThreeScaleBrowser
 from testsuite.ui.views.devel.login import LoginDevelView
+from testsuite.ui.webdriver import SeleniumDriver
 
 
 @pytest.fixture(scope="module")
-def custom_devel_auth0_login(browser, navigator, provider_account):
+def browser(request):
```
jsmolar, line 13: Isn't this copy-paste?

testsuite/tests/ui/conftest.py
```diff
@@ -347,6 +348,8 @@ def custom_rhsso_login(browser, navigator, threescale, request, testconfig):
     def _login(username, password, rhsso_username=None):
```
jsmolar, line 348: What is rhsso_username?

pehala (approved):

PR #91, opened 2022-01-03
Gather amp-version from apicast-staging trigger

pehala:
This should make sure the version is correct even in upgraded environments, as image triggers are present and updated in both template and operator upgrade/install.

### DIFF
--- a/testsuite/dynaconf_loader.py
+++ b/testsuite/dynaconf_loader.py
@@ -30,7 +30,6 @@ import yaml
 from openshift import OpenShiftPythonException
 from testsuite.openshift.client import OpenShiftClient
 
-
 identifier = "threescale"  # pylint: disable=invalid-name
 log = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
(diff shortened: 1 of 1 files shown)


testsuite/dynaconf_loader.py
```diff
@@ -53,17 +53,14 @@ def _guess_version(ocp):
...
-    lookup = _docker_image(ocp, "amp-system")
+    config = ocp.do_action("get", ["dc/apicast-staging", "-o", "yaml"], parse_output=True)
+    lookup = [trigger.imageChangeParams["from"]["name"].split(":", 1)[1]
+              for trigger in config.model.spec.triggers if trigger.type == "ImageChange"][0]
 
     try:
-        version = lookup["name"]
-        Version(version)
+        Version(lookup)
     except (KeyError, InvalidVersion):
```
mganisin, line 62: if original code expected KeyError this should probably handle IndexError and try should cover one line more

testsuite/dynaconf_loader.py
```diff
@@ -53,17 +52,13 @@ def _guess_version(ocp):
     """Attempt to determine version from amp-system imagestream"""
 
     version = None
-    lookup = _docker_image(ocp, "amp-system")
```
mganisin, line 56: `_docker_image` is now unused

testsuite/dynaconf_loader.py
```diff
@@ -53,17 +52,13 @@ def _guess_version(ocp):
...
-
     try:
-        version = lookup["name"]
+        config = ocp.do_action("get", ["dc/apicast-staging", "-o", "yaml"], parse_output=True)
```
mganisin, line 56: original code did derivate version from `amp-system`... ???

mganisin (changes requested):

testsuite/dynaconf_loader.py
pehala: amp-system can have a different image stream tag if used with Oracle so I changed it to apicast instead.

mganisin: apicast-staging can have different image stream tag if custom-policy is injected (yeah, less probable). apicast-production?

pehala: I don't think anyone is (and should be) manipulating apicast-staging deployment, however I will change it to apicast-production just to be safe

pehala: Changed

JaurbanRH (approved):

mganisin (approved):

PR #92, opened 2022-01-07
Add metrics tests for zync and zync-que

mdujava:


testsuite/tests/prometheus/zync/test_metrics.py
```diff
@@ -0,0 +1,74 @@
...
+
+@pytest.fixture(scope="module")
+def metrics_que(prometheus):
+    """Return all metrics from target defined of que."""
+    metrics = prometheus.get_metrics("que")
+    return metrics
+
+
+@pytest.fixture(scope="module")
+def metrics_zync(prometheus):
```
pehala, line 59: Is there any specific reason why this fixture is not parametrized as well?

mdujava: There is no reason, will fix it.

testsuite/tests/prometheus/zync/test_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+
+    return _metrics
+
+
+@pytest.mark.parametrize(("pod", "expected_metric"),
+                         [("que", x) for x in METRICS_QUE] + [("zync", x) for x in METRICS_ZYNC]
+                         )
+def test_metric_zync(request, expected_metric, pod):
+    """ Test metrics presense. """
+    actual_metrics = request.getfixturevalue("metrics")(pod)
```
pehala, line 66: Does the fact it calls for the pod for every metric slow it down?

mdujava: That's right, result is not cached.
It's equivalent to `actual_metrics = prometheus.get_metrics(pod)`.
So is significantly slower then version having separate fixture for each pod, which reuses the prometheus response.

What do you think, should I create 2 fixtures to cache metrics for each pod separately (to speed up tests) 
```
def metrics_que(prometheus):
    return prometheus.get_metrics("que")


def metrics_zync(prometheus):
    return prometheus.get_metrics("zync")
```

pehala: Yes, that would be the best

pehala (approved):

JaurbanRH (approved):

PR #99, opened 2022-01-19
Batching policy metrics

mdujava:


testsuite/tests/prometheus/apicast/selfmanaged/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+def require_openshift(testconfig):
+    """These tests require openshift available"""
+    if not weakget(testconfig)["openshift"]["servers"]["default"] % False:
+        warn_and_skip("All monitoring for selfmanaged apicast skipped due to missing openshift")
+
+
+@pytest.fixture(scope="module")
+def staging_gateway(request, gateway_kind, gateway_environment, gateway_options):
+    """Deploy self-managed template based apicast gateway."""
+    gw = gateway(kind=gateway_kind, staging=True, name=blame(request, "gw"), **gateway_options)
```
pehala, line 21: If you are using one gateway for all tests in this directory you can simply specify it here

testsuite/tests/prometheus/apicast/selfmanaged/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+    request.addfinalizer(gw.destroy)
+    gw.create()
+
+    if len(gateway_environment) > 0:
```
pehala, line 25: If you don't need an environment in subsequent tests you don't need to use it.

testsuite/tests/prometheus/apicast/selfmanaged/test_batching_caching_policy.py
```diff
@@ -0,0 +1,108 @@
...
+    """Adds policies to servies"""
+    service.proxy.list().policies.append(
+        rawobj.PolicyConfig("3scale_batcher", {"batch_report_seconds": BATCH_REPORT_SECONDS}))
+    return service
+
+
+@pytest.fixture(scope="module")
+def gateway_kind():
+    """Gateway class to use for tests"""
+    return SelfManagedApicast
```
pehala, line 30: Will this work also on OCP 3?

mdujava: no, as test expects to have Prometheus deployed by 3scale monitoring, which is only supported on ocp4.

@ficap @mijaros What do you think about test updating ConfigMap for a template-prometheus (adding new target)? I don't think it's a good idea, so I would keep this test ocp4 only.

pehala: In that case you can use OperatorApicast instead and limit this test on OCP4 (and skip it with ocp3)

testsuite/tests/prometheus/apicast/selfmanaged/conftest.py
mdujava: I removed thinks that I do not use, please check if my cleanup makes sense.

testsuite/tests/prometheus/apicast/selfmanaged/conftest.py
pehala: No, I mean getting rid of the gateway kind fixtures all together and using OperatorApicast as a kind instead

testsuite/tests/prometheus/apicast/selfmanaged/test_batching_caching_policy.py
mdujava: suggestion implemented.

JaurbanRH (approved):

pehala (approved):

PR #100, opened 2022-01-19
add labeling of ocp objects

mkudlej:
Add function for being able to add labels after creating objects.


mganisin (dismissed):

mkudlej:
@pehala I've found usecase for this PR. We need to be able to create secret and then to label it. This is not possible to do in one command `oc secret`. We need to create secret first and then label it.

pehala (approved):

PR #105, opened 2022-01-23
Mockserver instead of httpbin, requestbin and tinyproxy

mganisin:
This eliminates 3 services, replaced by one. Few other still remain (echo-api, go-httpbin), likely echo-api will stay there (as at least two services is needed), go-httbin implements http/2 and websockets, this is missing in mockserver now.

This PR is split into couple of commits, worth to review commit by commit. The change is relatively small with minimum of necessary changes to make the switch possible.

### DIFF
--- a/testsuite/tools.py
+++ b/testsuite/tools.py
@@ -1,9 +1,24 @@
 # pylint: disable=too-few-public-methods,broad-except
 
-"""This provides interface to get test environment tools
+"""
+This provides interface to get test environment tools
 
 Different sources can be used, e.g. if deployed in openshift tools can be gathered
-from openshift namespace/project."""
+from openshift namespace/project.
+
+Historically "symbolic" names used as config keys in settings use name of
+particular service e.g. 'httpbin'. This was never 100% accurate as the service
+shouldn't represent a specific implementation but rather available upstream
+API. It's better to think about these names as desired interface of the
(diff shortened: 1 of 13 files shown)


testsuite/tools.py
```diff
@@ -15,10 +15,10 @@
 
 _tr = {
     "echo_api": "echo-api+ssl",
-    "httpbin": "httpbin+ssl",
-    "httpbin_nossl": "httpbin",
+    "httpbin": "mockserver+ssl",
```
pehala, line 18: This just adds more confusion and removes any remaining transparency the tools had.

testsuite/tools.py
```diff
@@ -15,10 +15,10 @@
...
+    "httpbin_nossl": "mockserver",
     "httpbin_go": "go-httpbin+ssl",
-    "httpbin_service": "httpbin+svc",
+    "httpbin_service": "echo-api+svc:9292",
```
pehala, line 21: Why change it to echo-api when it is supposed to be an httpbin?

mganisin: https://github.com/3scale-qe/3scale-tests/pull/105/commits/761d8c7bb5a6368e40c77c0682d8673bb8807891

pehala (approved):

JaurbanRH (approved):

PR #107, opened 2022-01-24
Add test for GrafanaDashboard presence

mdujava:


testsuite/tests/grafana/test_dashboards.py
```diff
@@ -0,0 +1,43 @@
...
+     "sum(rate(rails_requests_total{namespace='$namespace',pod=~'zync-[a-z0-9]+-[a-z0-9]+'}[1m])) by (status)"),
+]
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("dashboard,expresion", DASHBOARDS)
+def test_dashboards(openshift, dashboard, expresion):
+    """check expresion to be inside of json definition of the dashboard"""
+    client = openshift()
+    ret = client.do_action('get', ['GrafanaDashboard', dashboard, '-ojson'])
```
pehala, line 34: `client.do_action('get', ['GrafanaDashboard', dashboard, '-ojson'], parse_output = True)` might solve the issue with big objects not being parsed as it returns APIObject, which can be used like `ret.model.spec.json` and shuld be already parsed

PR #110, opened 2022-01-25
Add test for GrafanaDashboard presence

mdujava:
This is new PR, renaming the branch closed original PR.

### DIFF
--- a/testsuite/tests/grafana/conftest.py
+++ b/testsuite/tests/grafana/conftest.py
@@ -1,0 +1,22 @@
+"""Check for openshift configuration"""
+from openshift import Missing
+from weakget import weakget
+import pytest
+
+from testsuite.utils import warn_and_skip
+
+
+@pytest.fixture(scope="module", autouse=True)
+def require_openshift(testconfig):
+    """These tests require openshift available"""
+    if not weakget(testconfig)["openshift"]["servers"]["default"] % False:
+        warn_and_skip("All monitoring for Grafana skipped due to missing openshift")
+
+
+@pytest.fixture(scope="module", autouse=True)
+def require_enabled_monitoring(openshift):
+    """Verify that monitoring is enabled in APIManager object"""
+    apimanager = openshift().api_manager
+
+    if apimanager.model.spec.monitoring is Missing or not apimanager.model.spec.monitoring.enabled:
+        warn_and_skip("Monitoring needs to be enabled in APIManager Object")
--- a/testsuite/tests/grafana/test_dashboards.py
+++ b/testsuite/tests/grafana/test_dashboards.py
@@ -1,0 +1,39 @@
+"""Tests for Grafana Dashboards definitions"""
+import pytest
+from packaging.version import Version  # noqa # pylint: disable=unused-import
+
+from testsuite import TESTED_VERSION  # noqa # pylint: disable=unused-import
+from testsuite.capabilities import Capability
+
+pytestmark = [
+        pytest.mark.skipif("TESTED_VERSION < Version('2.9')"),
+        pytest.mark.required_capabilities(Capability.OCP4),
+    ]
+
+DASHBOARDS = [
+    ("apicast-mainapp",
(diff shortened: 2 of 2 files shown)


testsuite/tests/grafana/test_dashboards.py
```diff
@@ -0,0 +1,37 @@
...
+        pytest.mark.skipif("TESTED_VERSION < Version('2.9')"),
+        pytest.mark.required_capabilities(Capability.OCP4),
+    ]
+
+# pylint: disable=line-too-long
+DASHBOARDS = [
+    ("apicast-mainapp",
+     "sum(rate(nginx_http_connections{namespace='$namespace', pod=~'apicast-$env.*'}[1m])) by (state)"),
+    ("apicast-services",
+     "sum(rate(upstream_status{namespace='$namespace', pod=~'apicast-$env.*', service_id='$service_id'}[1m])) by (status)"),  # noqa: E501
```
pehala, line 16: Not sure if it won't degrade the readability, but if you want to remove the long lines you can use this method of splitting it https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation.

testsuite/tests/grafana/test_dashboards.py
```diff
@@ -0,0 +1,37 @@
...
+    ("backend",
+     "sum(rate(apisonator_listener_response_codes{namespace=\\\"$namespace\\\",request_type=\\\"authorize\\\"}[1m]))"),  # noqa: E501
```
pehala, line 18: If you use `'` you can get rid of at least one \ for each escaped quote.

testsuite/tests/grafana/test_dashboards.py
```diff
@@ -0,0 +1,37 @@
...
+     "sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{namespace=\\\"$namespace\\\", pod=\\\"$pod\\\", container!=\\\"POD\\\", cluster=\\\"$cluster\\\"}) by (container)"),  # noqa: E501
+    ("system",
+     "sum(rate(rails_requests_total{namespace='$namespace',pod=~'system-app-[a-z0-9]+-[a-z0-9]+'}[1m])) by (status)"),
+    ("zync",
+     "sum(rate(rails_requests_total{namespace='$namespace',pod=~'zync-[a-z0-9]+-[a-z0-9]+'}[1m])) by (status)"),
+]
+# pylint: enable=line-too-long
+
+
+@pytest.mark.smoke
```
pehala, line 31: Smokes need to be run even without any additional configuration of the instance, is it really the case for grafana?

testsuite/tests/grafana/conftest.py
```diff
@@ -0,0 +1,13 @@
+"""Check for openshift configuration"""
+
+from weakget import weakget
+import pytest
+
+from testsuite.utils import warn_and_skip
+
+
+@pytest.fixture(scope="module", autouse=True)
+def require_openshift(testconfig):
```
pehala, line 10: Is this really the only thing that would skip this? Shouldnt there be a function checking if the grafana is even deployed and configured with the current instance?

testsuite/tests/grafana/test_dashboards.py
mdujava: I split the lines.

testsuite/tests/grafana/test_dashboards.py
mdujava: rewritten to use `'` in strings that contains `"`

testsuite/tests/grafana/test_dashboards.py
mdujava: this test is checking openshift object that are created by `monitoring > enabled : true` after openshift knows about grafana definitions.

I added check for the required configuration inside of `APIManager` object.

testsuite/tests/grafana/conftest.py
mdujava: added one more requirement mentioned in prev comment.

pehala (approved):

mganisin (approved):

PR #111, opened 2022-01-26
Custom Policy injection

palmieric:
This PR adds the automated test for the custom policy injection through secret via APIManager CRD.


testsuite/tests/apicast/operator/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+@pytest.fixture(scope="module")
+def secrets(openshift):
+    """Interface to return the dict-like Secrets
+    """
+    return Secrets(openshift())
+
+
+@pytest.fixture(scope="module")
+def delete_secret(secrets):
+    """Delete the secret containing the policy after execution
```
mganisin, line 19: why separate fixture and not part of create_custom_policy_secrets?

testsuite/tests/apicast/operator/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+    # We need to explicitly wait for the deployment being in starting state
+    # before waiting for it to be ready, the operator needs time to reconcile
+    # its state
+    wait_for_starting(apimanager)
+    wait_for_ready(apimanager)
+
+
+@pytest.fixture(scope="module")
+def service_settings(request) -> dict:
+    "dict of service settings to be used when service created"
```
mganisin, line 160: why is this and application copied here?

testsuite/tests/apicast/operator/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+    return {"name": blame(request, "svc")}
+
+
+# pylint: disable=too-many-arguments, too-many-instance-attributes
+@pytest.fixture(scope="module")
+def service(backends_mapping, custom_service, service_settings,
```
mganisin, line 166: maybe this can be done easier. as it is in testsuite/tests/apicast/policy/conftest.py
Actually it may make sense to move operator/ under testsuite/tests/apicast/policy/. thiughts?

mganisin (commented):
Overall nicely done

testsuite/tests/apicast/operator/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+from testsuite import rawobj
+from testsuite.utils import blame
+from testsuite.openshift.objects import Secrets
+
+
+@pytest.fixture(scope="module")
+def secrets(openshift):
+    """Interface to return the dict-like Secrets
+    """
+    return Secrets(openshift())
```
pehala, line 14: Woldnt openshift().secrets work also? In that way you dont need another fixture

testsuite/tests/apicast/operator/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def create_custom_policy_secret(secrets, delete_secret):
+    """Create an openshift secrets to use as custom policy based on https://github.com/3scale-qe/apicast-example-policy
+    """
+    initlua = "return require('example')"
+    apicastpolicyjson = """
+    {
```
pehala, line 32: I would put the policy into separate resource files as we would probably like to reuse it as much as possible.

testsuite/tests/apicast/operator/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+    return secrets
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(scope="module")
+def patch_apimanager(request, openshift, create_custom_policy_secret):
+    """Patch the apimanager CRD to add the custom policy added to apicast staging
+    """
+    apimanager = openshift().api_manager
+    # TODO: do it through apimanager.set_path but it looks not working at this time
```
pehala, line 116: `apimanager.modify_and_apply(lambda manager: manager.set_path("apicast/stagingSpec/customPolicies", <dict>"))` might work

testsuite/tests/apicast/operator/conftest.py
palmieric: I can move it to the root contest if you and @mganisin agree.

pehala: Not root conftest but rather separate files which you can then easily load in the code

mganisin: of course i was thinking about the same but i didnt comment at the end as i have no strong opinion. So far I had hard times to find a case where to reuse it.

pehala: There will be a separate testcase for APIcast operator, which doesn't have to be disruptive and be run more often.

I was also thinking about rewriting some of our custom policy tests so that they do not need a separate repository with code (might be not possible)

testsuite/tests/apicast/operator/conftest.py
palmieric: I agree

palmieric:
@mganisin @pehala applied all of your suggestions, please review again.

pehala (approved):

mganisin (approved):

PR #127, opened 2022-02-04
Add test to verify labels on operator pod

mdujava:
https://github.com/3scale/3scale-operator/pull/679 is redefining labels for 3scale operator. This MR is updating selector for operator pod.
New test is introduced that will verify label presence.


pehala (dismissed):

pehala (approved):

mganisin (approved):
For the record. I think there is a dange that this is too much specific test that can fail often due to changes in labels that may occur for any reason (like recently). Anyway, good to go.

PR #132, opened 2022-02-07
Extract Deployment related methods into separate module

pehala:
# Why?
* Currently `OpenShiftClient` is a big mess, if we want the same thing to works on both Deployment and DeploymentConfig we need one method for each of them
   *  For example,`_wait_for_deployment` and `wait_for_ready` - Both wait until the resource is ready, but one is for `DeploymentConfig` and one for `Deployment` (You can try to guess which one)
* Most of the methods are written to work only on `DeploymentConfig` even though they could easily work on `Deployment` too 

# Overview
* Adds Deployments class  
   * Enables all methods that worked on `DeploymentConfig` to also work on Kubernetes `Deployment`
   * Requires only a few methods to be overridden by `KubernetesDeployments` and `DeploymentConfigs` classes to work
* Change is not transparent
   * One of the problems I am trying to solve is that the naming is confusing and most methods aren't used more than once.
      * Only exception being `environ()` method, which is used extensively and which has transparency method.
      * However, if wanted, more of the methods could be made transparent with the default resource type being `DeploymentConfig`. I do not think that it is justified for most of them.
* This is a refactor MR, meaning I did not change functionality in any way.

# TODO
* Test that nothing broke

### DIFF
--- a/testsuite/gateways/apicast/operator.py
+++ b/testsuite/gateways/apicast/operator.py
@@ -124,7 +124,7 @@ class OperatorApicast(SelfManagedApicast):
 
     @property
     def deployment(self):
-        return f"apicast-{super().deployment}"
+        return self.openshift.deployment(f"deployment/apicast-{self.name}")
 
     @property
     def environ(self) -> Properties:
@@ -133,10 +133,7 @@ class OperatorApicast(SelfManagedApicast):
         return self._environ
 
     def reload(self):
-        self.openshift.do_action("delete", ["pod", "--force",
-                                            "--grace-period=0", "-l", f"deployment={self.deployment}"])
-        # pylint: disable=protected-access
-        self.openshift.wait_for_ready(self.deployment)
+        self.deployment.rollout()
 
     def create(self):
         apicast = APIcast.create_instance(
@@ -158,7 +155,7 @@ class OperatorApicast(SelfManagedApicast):
         # is created
         time.sleep(2)
         # pylint: disable=protected-access
-        self.openshift.wait_for_ready(self.deployment)
+        self.deployment.wait_for()
 
     def destroy(self):
         if self.apicast:
--- a/testsuite/gateways/apicast/selfmanaged.py
+++ b/testsuite/gateways/apicast/selfmanaged.py
@@ -8,6 +8,7 @@ from testsuite import utils
 from testsuite.capabilities import Capability
 from testsuite.gateways.apicast import AbstractApicast
 from testsuite.openshift.client import OpenShiftClient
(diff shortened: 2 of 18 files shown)


testsuite/gateways/apicast/operator.py
```diff
@@ -116,10 +116,7 @@ def environ(self) -> Properties:
         return self._environ
 
     def reload(self):
-        self.openshift.do_action("delete", ["pod", "--force",
-                                            "--grace-period=0", "-l", f"deployment={self.deployment}"])
-        # pylint: disable=protected-access
-        self.openshift.wait_for_ready(self.deployment)
+        self.openshift.deployments.rollout(self.deployment)
```
mganisin, line 119: these are not equal, while I am in favor of new suggestion this may be risky

testsuite/tests/prometheus/zync/test_annotations.py
```diff
@@ -21,7 +21,7 @@
 @pytest.fixture(params=['zync', 'zync-que'])
 def poda(request):
     """Return zync pod object."""
-    pod = openshift().get_pod(request.param).object()
+    pod = openshift().dc.get_pods(request.param).object()
```
mganisin, line 24: and what if zync will become deployment one day?

pehala: Then this one line will change

testsuite/gateways/apicast/operator.py
pehala: In which sense? I believe they are equal. The KubernetesDeployments rollout method should be just a copy of this.

pehala:
@mganisin 
* Changed the Deployments into Deployment, meaning all methods are called on the one respective object, instead of a class containing almost static methods.
   * `openshift.deployment("dc/test")` -> Returns DeploymentConfig
   * `openshift.deployment("deployment/test")`  -> KubernetesDeployment
* Gateway `deployment` method now correctly refers to an actual Deployment object - should end all the remaining confusion regarding this method

mganisin (dismissed):

testsuite/gateways/apicast/selfmanaged.py
```diff
@@ -68,9 +69,9 @@ def fits():   # pylint: disable=unused-argument
         return False
 
     @property
-    def deployment(self):
+    def deployment(self) -> Deployment:
         """Returns name of the deployment, by default it is just a name"""
```
mdujava, line 73: outdated doc string

mdujava (approved):

mdujava (approved):

mganisin (approved):

PR #133, opened 2022-02-08
Freeze pytest on 6.x version

pehala:
Pytest released 7.0 version on Friday and it broke pytest-cases (https://github.com/smarie/python-pytest-cases/issues/251) (It might have broken more but I didn't get that far). This is only a temporary workaround


pehala:
No longer needed, as pytest-cases is already fixed

PR #134, opened 2022-02-08
Add test for MessageBus variables in system-redis secret

mdujava:


testsuite/tests/system/test_system_redis_secret.py
```diff
@@ -0,0 +1,36 @@
...
+    """ return system-redis secret """
+    secret = openshift().secrets['system-redis']
+    return secret
+
+
+@pytest.mark.skipif("TESTED_VERSION >= Version('2.12')")
+@pytest.mark.parametrize("key", KEYS)
+def test_message_bus_secrets(system_redis_secret, key):
+    """ test of env variable presence """
+    assert system_redis_secret[key] is not None
```
pehala, line 28: Would using the `in` operator work? If yes, you could condense it into just one test

mdujava: `in` is not working, as only `__getitem__` is defined.

pehala (approved):

mganisin (approved):

PR #136, opened 2022-02-09
Make SelfManagedApicast more flexible

mganisin:
Primary idea behind this change is:
 - get ability to set instance of SelfManagedApicast explicitly
   (force option)

Secondary benefit is:
 - (CHANGE IN BEHAVIOR!) more dynamic selection
   (fallback to TemplateApicast actually as this is "always" available)

### DIFF
--- a/testsuite/gateways/apicast/selfmanaged.py
+++ b/testsuite/gateways/apicast/selfmanaged.py
@@ -1,16 +1,12 @@
 """Self-managed APIcast already deployed somewhere in OpenShift """
+import inspect
 import logging
-from typing import Dict, List
+from typing import Union, Type
 
-from threescale_api.resources import Service
-
-from testsuite import utils
 from testsuite.capabilities import Capability
-from testsuite.gateways.apicast import AbstractApicast
+from testsuite.gateways.gateways import Gateway, new_gateway
+from testsuite.gateways.apicast import AbstractApicast, OpenshiftApicast
 from testsuite.openshift.client import OpenShiftClient
-from testsuite.openshift.deployments import Deployment
-from testsuite.openshift.env import Properties
-from testsuite.openshift.objects import Routes
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


testsuite/gateways/apicast/selfmanaged.py
```diff
@@ -18,8 +15,16 @@ class NoSuitableApicastError(Exception):
     """Raised if no deployment method is found"""
 
 
+# pylint: disable=too-many-instance-attributes
 class SelfManagedApicast(AbstractApicast):
```
pehala, line 19: I would change this to inherit from `OpenShiftApicast` to not break typing. As the real class, it will return will be a subclass of `OpenShiftApicast`

mganisin: My thought is that this can be also `DockerApicast` one day. Thoughts? (not necessarily tomorrow or next week but generally speaking)

pehala: I was planning to rely on that assumptions while making TLS for OperatorApicast, so at least for now it would be better to have it bound to `OpenshiftApicast`, however we can reevaluate when we decide we want to implement `DockerApicast`.

testsuite/gateways/apicast/selfmanaged.py.orig
```diff
@@ -0,0 +1,301 @@
+"""Self-managed APIcast already deployed somewhere in OpenShift """
+import inspect
+import logging
+from typing import Union, Type
+
+from testsuite.capabilities import Capability
+from testsuite.gateways.gateways import Gateway, new_gateway
+from testsuite.gateways.apicast import AbstractApicast, OpenshiftApicast
+from testsuite.openshift.client import OpenShiftClient
+<<<<<<< HEAD
```
mdujava, line 10: added wrong file?

mganisin: Fixed!

mdujava (approved):

pehala (approved):

PR #144, opened 2022-02-17
Run make tests by files

pehala:
# Why
Problem with xdist I encounter is that with the default load balancer multiple tests in one file can be assigned to many processes which causes duplicate gateway (among others) initialization when it is not needed. This means that in many cases gateways are created much more often than they need to be causing unneeded load for cluster and 3scale (more unneeded service), increasing test times and chance for a random failure. This problem is only for tests that have expensive setups and there are more than one test case in the file.

# Solution
This setting simply groups tests by file before assigning them to processes which make removed this problem.

# TODO
* Test that nothing broke

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ all-is-package:
 
 test: ## Run test
 test pytest tests: pipenv
-	$(PYTEST) -n4 -m 'not flaky' $(flags) testsuite
+	$(PYTEST) -n4 -m 'not flaky' --dist loadfile $(flags) testsuite
 
 speedrun: ## Bigger than smoke faster than test
 speedrun: pipenv


mganisin (approved):
I disagree (== I will regret one day) :-) however I accept this change (at least for now).
In my opinion this have limited effect on load, but great potential to prolong test time under certain conditions.

This will decrease total number of deployments of standalone apicast but it has zero no effect on peeks, that is driven by number of xdist processes. Anyway it's fair to acknowledge that it may have some positive impact on load.

Contrary in case of long running tests (no to be confused with long running setup), e.g. those with time constraints, the execution will be significantly longer if the tests are in same file. Hopefully there isn't and won't be much tests like this (dangerous hope in my opinion)

mdujava (approved):

PR #165, opened 2022-03-01
Add gathering of apicast operator version

mdujava:
@mganisin I would have 2 questsions:
is the name TESTED_APICAST_VERSION ok, or would you suggest something different? APICAST_OPERATOR_VERSION ...
in get_apicast_operator: there is no clear label or other property that would specify subscription for Apicast operator, is it ok to expect that Apicast is the only operator in namespace?

### DIFF
--- a/config/settings.local.yaml.tpl
+++ b/config/settings.local.yaml.tpl
@@ -5,6 +5,7 @@
 development:  # default dynaconf env
   threescale:
     version: "${THREESCALE_VERSION}"
+    apicast_opearator_version: "${APICAST_OPERATOR_VERSION}"
     superdomain: ${THREESCALE_SUPERDOMAIN}
     admin:
       url: ${ADMIN_URL}
--- a/testsuite/__init__.py
+++ b/testsuite/__init__.py
@@ -74,5 +74,6 @@ else:
     os.environ["OPENSHIFT_CLIENT_PYTHON_DEFAULT_SKIP_TLS_VERIFY"] = "true"
 
 TESTED_VERSION = Version(str(settings["threescale"]["version"]))
+APICAST_OPERATOR_VERSION = Version(str(settings["threescale"]["apicast_operator_version"]))
 HTTP2 = settings.get("http2", False)
 ROOT_DIR = Path(os.path.abspath(__file__)).parent.parent
--- a/testsuite/dynaconf_loader.py
+++ b/testsuite/dynaconf_loader.py
@@ -24,6 +24,7 @@ import os
 import os.path
 
 from packaging.version import Version, InvalidVersion
+from weakget import weakget
 
 import yaml
 
@@ -57,6 +58,20 @@ def _guess_version(ocp):
         Version(version)
     except (ValueError, IndexError, OpenShiftPythonException, InvalidVersion):
         return ".".join(_testsuite_version().split(".")[:2])
+
+    return str(version)
+
+
+def _guess_apicast_operator_version(ocp):
+    """Attempt to determine version of apicast operator from subscription"""
+
+    version = None
+    try:
+        version = ocp.apicast_operator_subscription.object().model.status.installedCSV.split(".v")[1]
+        Version(version)
+    except (ValueError, IndexError, OpenShiftPythonException, InvalidVersion):
+        return "2022"  # returning value that is greater than any apicast version
+        # in result all tests for latest apicast operator will run
 
     return str(version)
 
@@ -109,6 +124,16 @@ def load(obj, env=None, silent=None, key=None):
             server_url=ocp_setup.get("server_url"),
             token=ocp_setup.get("token"))
 
+        apicast_ocp = ocp
+
+        apicast_section = weakget(obj)["threescale"]["gateway"]["OperatorApicast"]["openshift"] % None
+
+        if apicast_section:
+            apicast_ocp = OpenShiftClient(
+                project_name=apicast_section.get("project_name"),
+                server_url=apicast_section.get("server_url"),
+                token=apicast_section.get("token"))
+
         admin_url = _route2url(ocp.routes.for_service("system-provider")[0])
         admin_token = ocp.secrets["system-seed"]["ADMIN_ACCESS_TOKEN"].decode("utf-8")
         master_url = _route2url(ocp.routes.for_service("system-master")[0])
@@ -146,6 +171,7 @@ def load(obj, env=None, silent=None, key=None):
                         "server_url": ocp.do_action("whoami", ["--show-server"]).out().strip()}}},
             "threescale": {
                 "version": _guess_version(ocp),
+                "apicast_operator_version": _guess_apicast_operator_version(apicast_ocp),
                 "superdomain": superdomain,
                 "catalogsource": catalogsource,
                 "admin": {
--- a/testsuite/openshift/client.py
+++ b/testsuite/openshift/client.py
@@ -222,6 +222,19 @@ class OpenShiftClient:
 
         return self.select_resource("pods", narrow_function=select_operator)
 
+    @property
+    def apicast_operator_subscription(self):
+        """
+        Gets the selector for the apicast-operator subscription
+
+        :return: the subscription
+        """
+
+        def select_operator(subscription):
+            return subscription.model.spec.name == "apicast-operator"
+
+        return self.select_resource("subscriptions", narrow_function=select_operator)
+
     def select_resource(self,
                         resource: str,
                         labels: Optional[Dict[str, str]] = None,
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -159,6 +159,9 @@ def pytest_report_header(config):
     threescale = settings["threescale"]["admin"]["url"]
     version = settings["threescale"]["version"]
     catalogsource = weakget(settings)["threescale"]["catalogsource"] % "UNKNOWN"
+    apicast_project = \
+        weakget(settings)["threescale"]["gateway"]["OperatorApicast"]["openshift"]["project_name"] % "None"
+    apicast_version = settings["threescale"]["apicast_operator_version"]
     toolboximage = weakget(settings)["toolbox"]["podman_image"].split(':')[-1] % "UNKNOWN"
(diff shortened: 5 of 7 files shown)


testsuite/dynaconf_loader.py
```diff
@@ -109,6 +122,15 @@ def load(obj, env=None, silent=None, key=None):
             server_url=ocp_setup.get("server_url"),
             token=ocp_setup.get("token"))
 
+        apicast_ocp_setup = obj.get("gateway", {}).get("OperatorApicast", {}).get("openshift", {})\
```
pehala, line 125: weakget would probably work better

testsuite/openshift/client.py
```diff
@@ -213,6 +213,15 @@ def select_operator(apiobject):
...
         return self.select_resource("pods", narrow_function=select_operator)
 
+    def get_apicast_operator(self):
+        """
+        Gets the selector for the apicast operator
+
+        :return: the operator pod
+        """
+
+        return self.select_resource("subscriptions")
```
pehala, line 223: What if there will be multiple namespace operators? Will it still select only the apicast-operator?

testsuite/dynaconf_loader.py
```diff
@@ -61,6 +61,19 @@ def _guess_version(ocp):
     return str(version)
 
 
+def _guess_apicast_version(ocp):
+    """Attempt to determine version of apicast operator from subscription"""
+
+    version = None
+    try:
+        version = ocp.get_apicast_operator().object().status.installedCSV.split(".v")[1]
```
pehala, line 69: Will this return version 2.12 or 0.5.3?

mdujava: version of the operator not 3scale: `installedCSV: apicast-operator.v0.5.2`

testsuite/openshift/client.py
mdujava: That's why it's Draft, as this is for our current setup, where apicast-operator is only operator in namespace.

testsuite/dynaconf_loader.py
pehala: In that case why do you fallback to testsujte version which has different versioning?

testsuite/__init__.py
```diff
@@ -74,5 +74,6 @@ def _settimeout(self, timeout):
     os.environ["OPENSHIFT_CLIENT_PYTHON_DEFAULT_SKIP_TLS_VERIFY"] = "true"
 
 TESTED_VERSION = Version(str(settings["threescale"]["version"]))
+TESTED_APICAST_VERSION = Version(str(settings["threescale"]["gateway"]["OperatorApicast"]["version"]))
```
mganisin, line 77: this isn't good place. I suggest `["threescale"]["apicast_operator_version"]`

testsuite/tests/prometheus/apicast/selfmanaged/test_batching_caching_policy.py
```diff
@@ -5,11 +5,11 @@
 import pytest
 
 from packaging.version import Version  # noqa # pylint: disable=unused-import
-from testsuite import TESTED_VERSION, rawobj  # noqa # pylint: disable=unused-import
+from testsuite import TESTED_APICAST_VERSION, rawobj  # noqa # pylint: disable=unused-import
```
mganisin, line 8: APICAST_OPERATOR_VERSION sounds better. For example we could improve coverage of testing with different apicast version, in such case we could introduce also APICAST_TEMPLATE_VERSION. Better to be prepared in this case

mganisin (approved):

testsuite/dynaconf_loader.py
mdujava: addressed in new version

testsuite/openshift/client.py
mdujava: Added selector by the name

testsuite/__init__.py
mdujava: Moved to suggested path

testsuite/tests/prometheus/apicast/selfmanaged/test_batching_caching_policy.py
mdujava: renamed and added template version for now it's guessed from 3scale project

config/settings.local.yaml.tpl
```diff
@@ -5,6 +5,8 @@
 development:  # default dynaconf env
   threescale:
     version: "${THREESCALE_VERSION}"
+    apicast_opearator_version: "${APICAST_OPERATOR_VERSION}"
+    apicast_template_version: "${APICAST_VERSION}"
```
mganisin, line 9: I am afraid it is too early to include also template version. I was thinking it could be handy "one day", but we don't know yet how it will be needed. Better to remove it from this PR now and add it later when it will be really needed

testsuite/dynaconf_loader.py
```diff
@@ -109,6 +123,17 @@ def load(obj, env=None, silent=None, key=None):
             server_url=ocp_setup.get("server_url"),
             token=ocp_setup.get("token"))
 
+        apicast_ocp_setup = \
+            weakget(obj)["threescale"]["gateway"]["OperatorApicast"]["openshift"]["servers"]["default"] % ocp_setup
+
+        apicast_project = \
+            weakget(obj)["threescale"]["gateway"]["OperatorApicast"]["openshift"]["project_name"] % "apicast-ga"
```
mdujava, line 130: @pehala 
Is this the right place to specify openshift server for apicast operator?
This should be in default configs set up to `apicast-ga` or `apicast-operator` so I can change this to normal lookup?

testsuite/dynaconf_loader.py
```diff
@@ -109,6 +123,17 @@ def load(obj, env=None, silent=None, key=None):
             server_url=ocp_setup.get("server_url"),
             token=ocp_setup.get("token"))
 
+        apicast_ocp_setup = \
+            weakget(obj)["threescale"]["gateway"]["OperatorApicast"]["openshift"]["servers"]["default"] % ocp_setup
```
mganisin, line 127: this is nice, just wanted to ask @pehala to check this, is this correct "path" to potential server definition (I think we dont need/use it now). I would probably say "no" as it looks to deep to me.

testsuite/openshift/client.py
```diff
@@ -213,6 +213,18 @@ def select_operator(apiobject):
 
         return self.select_resource("pods", narrow_function=select_operator)
 
+    def get_apicast_operator(self):
```
mganisin, line 216: it would be nice if this was property for now

testsuite/openshift/client.py
```diff
@@ -213,6 +213,18 @@ def select_operator(apiobject):
...
+        """
+        Gets the selector for the apicast operator
```
mganisin, line 218: this says "selector", return description says "pod"... not sure what it returns :-)

mdujava: fixed

testsuite/openshift/client.py
mdujava: fixed.

testsuite/dynaconf_loader.py
pehala: The best way would be probably something like:
```
client = ocp
apicast_section = weakget(obj)["threescale"]["gateway"]["OperatorApicast"]["openshift"] % None
if apicast_section:
  client = OpenshiftClient(**apicast_section)
```

testsuite/dynaconf_loader.py
```diff
@@ -109,6 +123,16 @@ def load(obj, env=None, silent=None, key=None):
             server_url=ocp_setup.get("server_url"),
             token=ocp_setup.get("token"))
 
+        apicast_ocp = ocp
+
+        apicast_section = weakget(obj)["threescale"]["gateway"]["OperatorApicast"]["openshift"] % None
+
+        if apicast_section:
+            apicast_ocp = OpenShiftClient(
```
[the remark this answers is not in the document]
mdujava: Oh, sorry, I forget to justify it, There is problem with it as this section contains "kind", and I was not able to get it running.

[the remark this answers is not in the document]
pehala: Makes sense, will this also work if some of the values are missing? The only required value is project_name

testsuite/dynaconf_loader.py
```diff
@@ -109,6 +123,16 @@ def load(obj, env=None, silent=None, key=None):
...
+                project_name=apicast_section.get("project"),
```
pehala, line 132: project_name

testsuite/dynaconf_loader.py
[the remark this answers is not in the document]
mdujava: yes, it would resolve to None and it would chose default server from openshift section or get it from logged in openshift in cli

testsuite/tests/conftest.py
```diff
@@ -159,6 +159,8 @@ def pytest_report_header(config):
     threescale = settings["threescale"]["admin"]["url"]
     version = settings["threescale"]["version"]
     catalogsource = weakget(settings)["threescale"]["catalogsource"] % "UNKNOWN"
+    apicast_project = settings["threescale"]["gateway"]["OperatorApicast"]["openshift"]["project_name"]
```
pehala, line 162: This might fail, I would use weakget just to be sure

mdujava: fixed.

pehala (approved):

testsuite/openshift/client.py
```diff
@@ -222,6 +222,19 @@ def select_operator(apiobject):
 
         return self.select_resource("pods", narrow_function=select_operator)
 
+    @property
+    def get_apicast_operator_subscription(self):
```
mganisin, line 226: properties don't have `get_`

pehala (approved):

testsuite/openshift/client.py
mdujava: fixed.

pehala (approved):

mganisin (approved):

PR #180, opened 2022-03-10
Added a full-page (multiple screenshots) screenshot functionality

dlaso99:
This PR adds the code that generates multiple images regarding the height of the website. Because of some CSS related stuff, the HTML of the website had to be altered (2 styles were added).

At first, I wanted to create 1 big screenshot of the website. Problem is that even the existing libraries are having trouble to make a perfect fullscreen image of the website when using selenium (Google chrome for some reason does not return correct values about the height of the website (atleast to me it is not) and Firefox has integrated a fullscreen capturing). 

To be a bit more specific, what happens is that google chrome (chromedriver) returns a height of the website viewport equal to `839px`, but makes a screenshot of a height of `891px`. I have no idea why this is happening, it works perfectly on firefox, but not on Chrome. 

In order to merge 2 png files, we would need to get dimensions of the screenshot and create a metric that will allow us to create the whole image (I have tried to do something simillar a while ago but I was not successful. Therefore for now, the website split into the multiple images with offset (so that nothing gets lost) will hopefully be enough.

### DIFF
--- a/testsuite/tests/ui/conftest.py
+++ b/testsuite/tests/ui/conftest.py
@@ -288,8 +288,7 @@ def pytest_exception_interact(node, call, report):
         browser = node.funcargs.get("browser")
         if not browser:
             return
-        screenshot = os.path.join(get_resultsdir_path(node), "failed-test-screenshot.png")
-        browser.selenium.save_screenshot(screenshot)
+        fullpage_screenshot(driver=browser.selenium, file_path=get_resultsdir_path(node))
 
 
 def get_resultsdir_path(node):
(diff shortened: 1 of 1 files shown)


mganisin (approved):

testsuite/tests/ui/conftest.py
```diff
@@ -322,6 +321,29 @@ def get_resultsdir_path(node):
     return path
 
 
+def fullpage_screenshot(driver, file_path):
+    """
+        A full-page screenshot function. It scroll the website and screenshots it.
+        - Creates multiple files
+        - Screenshots are made only vertically (on Y axis)
+    """
+    # Removal of the height: 100% style, that disables scroll.
```
JaurbanRH, line 330: As @dhlavac mentioned if this block of code throws some error the whole test will fail even the test run passed. Add try block to prevent this from happening.

dlaso99: The question is then ... do we want to let the tester know that screenshots failed as well ?

dlaso99:
Reworked and moved to -> https://github.com/3scale-qe/3scale-tests/pull/193

PR #184, opened 2022-03-15
Fix wrong response codes in some tests with WASM

azgabur:


testsuite/tests/system/mapping/test_mapping_rules.py
```diff
@@ -65,15 +65,15 @@ def client(application, api_client):
 
 
 @pytest.fixture(scope="session")
-def error_status_code(staging_gateway):
+def error_status_code_404(staging_gateway):
```
pehala, line 68: I would not change this retrospectively

testsuite/tests/apicast/test_limit_exceeded_metric.py
```diff
@@ -29,14 +30,23 @@ def application(application):
     return application
 
 
-def test_limit_exceeded(api_client):
+@pytest.fixture(scope="session")
+def error_status_code_429(staging_gateway):
```
pehala, line 34: `limit_reached_status_code` or something similar would be probably better

pehala (changes requested):

testsuite/tests/apicast/test_limit_exceeded_metric.py
azgabur: Renamed, is it better now?

mganisin (approved):

PR #194, opened 2022-03-22
Introduce async into the testsuite

pehala:
Replaces our custom event loops with async tests and methods instead. This PR should not change the functionality.


mganisin (approved):
two files seem to still conflict with the other asyncio PR. Of course first gets in first wins and the second will need an update, so I am going to approve this anyway

mdujava (dismissed):

mganisin (approved):

JaurbanRH (approved):

PR #204, opened 2022-03-24
Add UI part for handling users permission

dhlavac:


testsuite/ui/widgets/__init__.py
```diff
@@ -17,52 +17,91 @@
...
+
+
+# pylint: disable=abstract-method
+# Widget contains fill method which raise not implemented exception if widget is not fillable but pylint detect it as
+# an abstract method
+class CheckBoxGroup(GenericLocatorWidget):
+    """
+    CheckBox group of 3scale pages
+    :param ol_class: Set if Checkbox group identifier is in ol level of group.
+                     e.g. fieldse/ol[class=ServiceList]
```
jsmolar, line 52: you missed "t" in fieldset

jsmolar: Also, I would use  common naming: Set if Checkbox group is identified by attribute in `ol `tag

testsuite/ui/widgets/__init__.py
```diff
@@ -17,52 +17,91 @@
...
+    OPTIONS_SECTION = './fieldset/ol'
+    OPTIONS_SECTION_OL_CLASS = './/fieldset/ol[contains(@class, "{}")]'
+    OPTIONS_SECTION_FIELDSET_NAME = './/fieldset[contains(@name, "{}")]/ol'
+
+    OPTIONS = './li'
+    OPTIONS_INPUT = OPTIONS + '/label/input'
+    OPTIONS_BY_ID = OPTIONS_INPUT + '[@id="{}"]'
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, parent=None, locator=None, ol_class=None, field_set_name=None, logger=None):
```
jsmolar, line 64: Please unify naming of `ol_class` and `field_set_name` so it is obvious that they are used for the identification of the group

testsuite/ui/widgets/__init__.py
```diff
@@ -17,52 +17,91 @@
...
+        super().__init__(parent, locator, logger)
+        if ol_class:
+            self.options_section = self.browser.element(self.OPTIONS_SECTION_OL_CLASS.format(ol_class))
```
jsmolar, line 67: `ol_class` and `field_set_name` should be saved in locator of the widget.. At least that is common usage in widgetastic

jsmolar: If you save it into locator you save some code. Example from your code:
`self.browser.element(self.OPTIONS_BY_ID.format(option), parent=self.options_section)`
with locator:
`self.browser.element(self.OPTIONS_BY_ID.format(option))`

testsuite/ui/widgets/__init__.py
```diff
@@ -17,52 +17,91 @@
...
+            self.options_section = self.browser.element(self.OPTIONS_SECTION_FIELDSET_NAME.format(field_set_name))
+        else:
+            self.options_section = self.browser.element(self.OPTIONS_SECTION)
+
+    def is_checked(self, option: str):
+        """Detect if checkbox is already checked"""
+        element = self.browser.element(self.OPTIONS_BY_ID.format(option), parent=self.options_section)
+        return element.is_selected()
+
+    def check(self, options: list):
```
jsmolar, line 78: this method should be consistent. It should uncheck all options before checking new ones

testsuite/ui/widgets/__init__.py
dhlavac: Changed to locator form

testsuite/ui/widgets/__init__.py
dlaso99: @jsmolar  what about the moment when I want to select an additional option? Here is an example of what I mean:
```python
(pseudocode)
check(["blue", "red"])                        # adds the whole list of options
// do some asserts
check("green")                                # is not supported rn
check(["blue", "red"] + ["green"])           # what i have to do right now
```
The current implementation requires me to send the whole list of options that I want to have checked, but what if I want to keep the old options and just check a single check ?

testsuite/ui/widgets/__init__.py
```diff
@@ -17,52 +17,93 @@
...
+
+    OPTIONS = './li'
+    OPTIONS_INPUT = OPTIONS + '/label/input'
+    OPTIONS_BY_ID = OPTIONS_INPUT + '[@id="{}"]'
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, parent=None, locator=None, ol_identifier=None, field_set_identifier=None, logger=None):
+        super().__init__(parent, locator, logger)
+        if ol_identifier:
+            self.locator = locator + self.OPTIONS_SECTION.format(ol_identifier)
```
jsmolar, line 67: le dot.

testsuite/ui/views/admin/settings/webhooks.py
```diff
@@ -15,10 +15,10 @@ class WebhooksView(BaseSettingsView):
     path_pattern = '/p/admin/webhooks/edit'
     webhook_active = GenericLocatorWidget('//*[@id="web_hook_active"]')
     webhook_provider = GenericLocatorWidget('//*[@id="web_hook_provider_actions"]')
-    accounts = CheckBoxGroup('//*[@name="Accounts"]')
-    users = CheckBoxGroup('//*[@name="Users"]')
-    applications = CheckBoxGroup('//*[@name="Applications"]')
-    keys = CheckBoxGroup('//*[@name="Keys"]')
+    accounts = CheckBoxGroup('//tr/td', field_set_identifier="Accounts")
```
jsmolar, line 18: Can't we use here something better than `//tr/td` ? Like `//*[@id="new_web_hook"]/fieldset[@class="inputs"]` Not like this but simmilar

JaurbanRH (approved):

dlaso99 (approved):
lgtm!

PR #228, opened 2022-04-27
extra dir to separate external tools

mganisin:


Makefile
```diff
@@ -142,11 +142,12 @@ dist: pipenv fetch-tools
 	-[ -n "$$NOSWITCH" ] || git checkout -
 
 fetch-tools:
-	-rm -Rf testsuite-tools/
-	-curl $(fetch_tools) | tar -xz
+	-rm -Rf ext/testsuite-tools/
+	-mkdir -p ext/
+	-curl $(fetch_tools) | tar -C ext/ -xz
```
mdujava, line 147: I don't think `curl |tar` return code should be ignored

Makefile
```diff
@@ -142,11 +142,12 @@ dist: pipenv fetch-tools
...
 
 tools:
```
mdujava, line 149: I would add dependency on `ext/testsuite-tools/run.sh` or `fetch-tools` so it does not fail with file not found?

mganisin: No. `fetch-tools` is build action (called during container build) while `tools` is runtime action (as tools url doesn't have to be available at runtime), they are intentionally split

Makefile
mganisin: and I think it should :-)
It is ignored because it is ran everytime when building container, but tools url doesn't have to be available every time. Yes, whole tools target works just by "accident" and it isn't guaranteed. That's because it isn't guaranteed and purely optional. The rule is simple `make fetch-tools` (which should not be called explicitly) should never fail - should never block build process, rather build container with `make tools` not working

pehala (approved):

JaurbanRH (approved):

PR #234, opened 2022-04-28
Read resource by name in UI tests with default backoff

dhlavac:
Added default backoff to UI test when the resource is created in UI but may not be processed by server before API call

### DIFF
--- a/testsuite/tests/ui/apiap/test_backend.py
+++ b/testsuite/tests/ui/apiap/test_backend.py
@@ -1,6 +1,7 @@
 """Rewrite of spec/ui_specs/api_as_a_product/create_backend_spec.rb.rb"""
 import pytest
 
+from testsuite import resilient
 from testsuite.utils import blame
 from testsuite.ui.views.admin.backend.backend import BackendEditView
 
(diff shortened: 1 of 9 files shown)


testsuite/tests/ui/apiap/test_backend.py
```diff
@@ -55,7 +56,7 @@ def test_delete_backend(login, navigator, threescale, custom_backend):
     backend = custom_backend(autoclean=False)
     edit = navigator.navigate(BackendEditView, backend=backend)
     edit.delete()
-    backend = threescale.backends.read_by_name(backend.entity_name)
+    backend = resilient.resource_read_by_name(threescale.backends, backend.entity_name)
```
pehala, line 59: Shouldn't the back off be reverted here? I thought we are testing that the backend no longer exist.

dhlavac: @pehala should be reverted. In case of deletion test basic API call is used

mganisin (approved):

JaurbanRH (approved):

PR #237, opened 2022-05-02
Removed unnecessary xfailed and simplified test_account_search

jsmolar:


testsuite/tests/ui/test_account_search.py
```diff
@@ -43,18 +42,7 @@ def test_search_account(login, navigator, custom_ui_account, ui_application, req
...
-    assert next(accounts.table.rows())[1].text == org_name
-    assert next(accounts.table.rows())[6].text == "Approved"
-
-    accounts.search(app["name"])
-    assert next(accounts.table.rows())[1].text == org_name
-    assert next(accounts.table.rows())[6].text == "Approved"
+    for key in [org_name, username, email, app["name"]]:
+        accounts.search(key)
+        assert accounts.table.row()[1].text == org_name
+        assert accounts.table.row()[5].text == "Approved"
```
pehala, line 48: This seems like a breaking change, as before you searched in column 6 and now you search in 5, is that expected?

jsmolar: Yes, the 6 column has an empty string now. Don't know why.

JaurbanRH (approved):

mganisin (approved):

PR #249, opened 2022-05-05
Refactor stripe and braintree payment asserts

jsmolar:


testsuite/billing.py
```diff
@@ -23,7 +24,7 @@ def read_customer_by_account(account):
         return [x for x in stripe.Customer.list() if
                 str(account.entity_id) in x["metadata"].get("3scale_account_reference", [])][0]
 
-    def invoice_assert(self, account, invoice):
+    def assert_payment(self, account, invoice):
```
pehala, line 27: Why does this have one more parameter than the same function in Braintree?

jsmolar: @JaurbanRH ^^

JaurbanRH: you can change it to something like this:
`    @staticmethod
    def read_payment(order_id):
        """Read Stripe payment"""
        return [x for x in stripe.PaymentIntent.list() if x['metadata']['order_id'] == order_id][0]

    def invoice_assert(self, invoice):
        """Compare 3scale and Stripe invoices"""
        stripe_invoice = self.read_payment(str(invoice['id']))
        currency = stripe_invoice['currency']
        # Stripe amount is Integer, e.g. 10,50$ is as 1050, so we need to divide it by 100 to get wanted cost
        cost = stripe_invoice['amount'] / 100
        # Compare 3scale invoice values and Stripe invoice values and check whether Stripe invoice is marked as paid
        return currency == invoice['currency'].lower() \
            and cost == invoice['cost'] \
            and stripe_invoice['charges']['data'][0]['paid']`

dhlavac (approved):

PR #274, opened 2022-05-19
Ensure proper cleanup for all WASM resources

pehala:


testsuite/gateways/wasm/__init__.py
```diff
@@ -65,7 +65,10 @@ def create(self):
         })
 
     def destroy(self):
-        self.mesh.delete_app(self.label, resources="serviceentry")
+        self.mesh.delete_app(self.label, resources="serviceentry,destinationrule")
+        for extension in self.extensions:
+            extension.delete()
+        self.extensions = []
```
azgabur, line 71: `dict()` or just `{}`

testsuite/gateways/wasm/__init__.py
```diff
@@ -65,7 +65,10 @@ def create(self):
         })
 
     def destroy(self):
-        self.mesh.delete_app(self.label, resources="serviceentry")
+        self.mesh.delete_app(self.label, resources="serviceentry,destinationrule")
+        for extension in self.extensions:
+            extension.delete()
```
azgabur, line 70: Is this required? Before this commit the cleanup of openshift objects made in extension.py was done correctly as far as I know.

azgabur (changes requested):

testsuite/gateways/wasm/__init__.py
pehala: Theoretically, no, I just wanted to make sure it gets deleted. I removed it for now and if we see some cases where it doesn't clean up properly I will re-add it

azgabur (approved):

mganisin (approved):

PR #289, opened 2022-06-08
Indirect fix for test_account_search

DaniilRomanov:
This correction concerns mostly test_account_search. Sphinx search server with
default settings splits test strings generated by blame() in parts (originally
separated by '-' symbol). If there is at least one part less than 3 characters
then search in this test fails. The correction fixes test string generation in
the part where previously could be less 3 characters.


dhlavac:
I think that this is a valid issue, so we shouldn't correct the tests. Customers can use suffix for country codes(which are mostly 2 characters) and in that case, the whole search doesn't work.

testsuite/utils.py
```diff
@@ -62,7 +68,13 @@ def blame(request: 'FixtureRequest', name: str, tail: int = 3) -> str:
     if "." in context:
         context = context.split(".")[0]
 
-    return randomize(f"{name[:8]}-{_whoami()[:8]}-{context[:9]}", tail=tail)
+    # When Sphinx search works with strings generated by blame(), parts separated by '-' symbol
```
mganisin, line 71: My understanding is that described behavior is unexpected bug, we should not mitigate it thus, maybe we could rather mark affected test as xfail with reference to relevant bug.

On the other hand change above in `whoami` is nice one worth to merge, my suggestion is to rename commit remove changes in blame and repush it just with changes in `whoami`

mganisin (changes requested):

testsuite/utils.py
DaniilRomanov: Made new PR: https://github.com/3scale-qe/3scale-tests/pull/300

PR #297, opened 2022-06-20
Migrate image to centos-stream

pehala:
- Migrate to centos-stream image
  - The old base image is based on centos7 and does support newer python images 
- Update python to 3.9
- ~~Remove explicit CA links~~
   - ~~Should be supplied through secret~~
- Remove unnecessary dependencies (`docker-client`, `openssh-client`)
   -  Docker is unused and OpenSSH shouldn't be needed as paramiko implements the protocol itself (@mdujava please verify)
- ~~Venv is now created inside workdir~~
- ~~Image is not buildable outside of VPN~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/centos7/python-38-centos7
+FROM quay.io/centos/centos:stream
 LABEL description="Run 3scale integration tests \
 Default ENTRYPOINT: 'make' and CMD: 'smoke' \
 Bind dynaconf settings to /opt/secrets.yaml \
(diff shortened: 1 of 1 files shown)


mdujava (dismissed):
toolbox/paramiko works just fine

mganisin (changes requested):

pehala:
@mganisin Reworked based on our discussions.

mdujava (dismissed):

Dockerfile
```diff
@@ -15,14 +15,15 @@ ADD $cacert /etc/pki/ca-trust/source/anchors
 ADD https://gist.githubusercontent.com/mijaros/c9c9ed016ce9985d96c6c5c3b35b4050/raw/66587720883554b03a4c24875fa47442db231a51/ca.pem /etc/pki/ca-trust/source/anchors
 RUN update-ca-trust
 
+RUN useradd --no-log-init -u 1001 -G root -r -m default
```
mganisin, line 18: this should be `-g root` and I would really like to see also `-r` removed (the latter is mild, however unless there is good reasoning I'd prefer to make sure `$HOME` exists; actually there is reference to `$HOME later in Dockerfile, likely to eradicate pipenv cache, so it is highly desirable to ensure it is properly set with correct access permissions)

mganisin (approved):

mdujava (approved):

PR #298, opened 2022-06-20
Add cluster-wide project to be searched for apicast-operator version

mdujava:


testsuite/dynaconf_loader.py
```diff
@@ -153,6 +153,13 @@ def load(obj, env=None, silent=None, key=None):
 
         apicast_ocp = _apicast_ocp(ocp, obj)
 
+        apicast_operator_version = _guess_apicast_operator_version(apicast_ocp)
+        if apicast_operator_version == "50035350":
```
pehala, line 157: If you want to use it in a condition, you should make the "default" version a constant, also why not add it into the original function?

pehala (approved):

mganisin (approved):

PR #300, opened 2022-06-21
Fix for _whoami() function in testsuite/utils.py

DaniilRomanov:
In some environments os.getlogin() returns an error, for example in a shell
inside Docker container, so it was replaced by os.getuid() and user name
resolution.

### DIFF
--- a/testsuite/utils.py
+++ b/testsuite/utils.py
@@ -38,7 +38,12 @@ def _whoami():
     # want to catch broad exception and fallback at any circumstance
     # pylint: disable=broad-except
     except Exception:
-        return str(os.getuid())
+        try:
+            # this expression is an equivalent of os.getlogin() for some environments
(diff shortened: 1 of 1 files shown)


testsuite/utils.py
```diff
@@ -38,7 +38,12 @@ def _whoami():
     # want to catch broad exception and fallback at any circumstance
     # pylint: disable=broad-except
     except Exception:
-        return str(os.getuid())
+        try:
+            # this expression is an equivalent of os.getlogin() for some environments
+            return pwd.getpwuid(os.getuid()).pw_name
```
pehala, line 43: It seems you are missing an import

mganisin:
@DaniilRomanov would you mind to update this PR in accordance to failed check?

mganisin:
inactive

PR #304, opened 2022-07-04
Handle apicast operator version in case of unknown namespce

mganisin:


mdujava (dismissed):

testsuite/dynaconf_loader.py
```diff
@@ -65,6 +65,8 @@ def _guess_version(ocp):
 def _guess_apicast_operator_version(ocp, settings):
     """Attempt to determine version of apicast operator from subscription"""
 
+    if not ocp or not ocp.project_name:
+        return 0
```
pehala, line 69: Why is this inconsistent with the other value we use as a default?

mdujava (commented):
If there is no project_name, we still want to check 'openshift-operators' namespace.

testsuite/dynaconf_loader.py
mganisin: apicast-operator is unkown or not present in this case, neither devel version (what is the other number). "0" may not be smartest value, implies some handling that doesn't have to be straightforward, anyway it is compatible with version format and "says" exactly what version of apicast-operator is present. other possible value is "99999999", it looks more confusing and less descriptive.

pehala (approved):

mdujava (approved):

pehala (approved):

PR #305, opened 2022-07-04
Ensure correct openshift and namespace is used when selecting gateway kind

mganisin:


mdujava (approved):

testsuite/gateways/apicast/selfmanaged.py
```diff
@@ -47,12 +49,23 @@ def __new__(cls, staging: bool, openshift: OpenShiftClient, settings,
...
                         break
+
+            def _openshift(subclass):
+                """get right openshift for a subclass"""
+                custom = weakget(settings)[subclass.__name__]["openshift"] % None
+                if not custom:
+                    return openshift
+                if isinstance(custom, OpenShiftClient):
+                    return custom
+                return OpenShiftClient(
```
pehala, line 60: I don't understand why this is needed

mganisin: Default openshift is passed to `SelfManagerApicast` that is not openshift that for example `OperatorApicast` should have as it can have (and it really does have) custom openshift settings. Direct glitch is that `OpenshiftClient.project_exists` returns wrong value (3scale namespace vs. apicast-operator namespace).

pehala: Yeah, but why do you create OpenShiftClient directly, if configured properly it should always be an OpenShiftClient already.

mganisin: I don't think so, if `SelManagedApicast` is used, `OperatorApicast` section shouldn't be touched and processed. This part is happening here

mganisin: Contrary I can't imagine openshift instance could be ever here, anyway one `if` doesn't hurt (at least here, at least for now)

pehala (approved):

PR #306, opened 2022-07-04
resilient tenant

mganisin:
- Set ssl verification for tenant/wait_tenant_ready
- Reset extra waiting period
- Implement resilient accsounts.create to handle 409

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -333,7 +333,7 @@ def threescale(testconfig):
         testconfig["threescale"]["admin"]["url"],
         testconfig["threescale"]["admin"]["token"],
         ssl_verify=testconfig["ssl_verify"],
-        wait=16
+        wait=0
     )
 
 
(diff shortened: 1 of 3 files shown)


mdujava (approved):

testsuite/tests/conftest.py
```diff
@@ -333,7 +333,7 @@ def threescale(testconfig):
         testconfig["threescale"]["admin"]["url"],
         testconfig["threescale"]["admin"]["token"],
         ssl_verify=testconfig["ssl_verify"],
-        wait=16
+        wait=0
```
pehala, line 336: Do we really need this parameter? Cannot we just omit it?

mganisin: No, no param means no waiting, `wait=0` means "wait until condition is met" `wait=16` means "wait next 16 seconds after condition is met". `wait=0` behavior is desirable and no waiting is reasonable as library default, therefore this custom value here

pehala (approved):

PR #312, opened 2022-07-08
Fix has_apicast_operator check

mganisin:
openshift "always" returns something, object should be checked after
they are mined from openshift reply.


mdujava (approved):

pehala (approved):
Not sure if the object() method shouldn't be called within context but we can fix it afterwards

PR #314, opened 2022-07-11
Upgraded Selenium to version 4

jsmolar:


Pipfile
```diff
@@ -31,7 +31,6 @@ python-keycloak = "*"
 backoff = "*"
 websocket_client = "*"
 httpx = {version = "*", extras = ["http2"]}
-msedge-selenium-tools = "*"
 webdriver-manager = "*"
 widgetastic-core = "*"
```
pehala, line 35: If you want to assure that Selenium is bigger than 4.0 (which is probably needed since you updated the code) you should explicitly force it here

pehala (approved):

JaurbanRH (approved):

PR #315, opened 2022-07-13
Change dc wait_for to use rollout status

mdujava:
This will wait for new Pods to be created and started up; and old pods to be deleted.

Original code might return prior to new pods being created, so DeploymentConfig would still run with old config.


pehala (approved):

mganisin:
If someone (@pehala ) would remember why oc rollout status might not be enough....
Maybe both checks together could be better...
Anyway....

PR #335, opened 2022-07-24
Change report header structure

mganisin:
Simplify report header rendering. Add more to junit testsuite
properties. Use dynaconf env name to report values. Use pytest-metadata.
Print also env settings values.

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -122,104 +122,78 @@ def pytest_collection_modifyitems(session, config, items):
             item.user_properties.append(("issue-id", issue_id))
 
 
-def _oc_3scale_project():
-    """Just a one liner to not repeat really long line"""
-
-    return weakget(settings)["openshift"]["projects"]["threescale"]["name"] % "UNKNOWN"
-
-
-# Currently this doesn't work with xdist
 # https://github.com/pytest-dev/pytest/issues/7767
-# Therefore _global_property hack is present below
-# **AND** therefore @fixture commented out to avoid properties
-# doubled on single execution. Once xdist issue will be fixed
-# this will be updated
-# @pytest.fixture(scope="session", autouse=True)
-def testsuite_properties(record_testsuite_property):
-    """Add custom testsuite properties to junit"""
-
-    title = os.environ.get("JOB_NAME", "Ad-hoc").split()[0]
-    runid = f"{title} {_oc_3scale_project()} {settings['threescale']['version']}"
-    projectid = weakget(settings)["reporting"]["testsuite_properties"]["polarion_project_id"] % "None"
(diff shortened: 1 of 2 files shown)


mdujava (approved):

testsuite/tests/conftest.py
```diff
@@ -122,104 +122,78 @@ def pytest_collection_modifyitems(session, config, items):
...
+            toolboxversion = toolbox.run_cmd("-v")['stdout'].strip()
+        except BoxKeyError:
+            warnings.warn("Toolbox executioner is not configured")
 
-        if toolboximage != "UNKNOWN":
-            try:
-                toolboxversion = toolbox.run_cmd("-v")['stdout'].strip()
-            except BoxKeyError:
-                warnings.warn("Toolbox executioner is not configured")
+    metadata["toolbox-version"] = toolboxversion
```
pehala, line 183: I am not sure if we want to include this if toolbox tests aren't run

testsuite/tests/conftest.py
```diff
@@ -122,104 +122,78 @@ def pytest_collection_modifyitems(session, config, items):
...
-        header.append(f"testsuite: toolbox image tag = {toolboximage}")
 
+# pylint: disable=protected-access
+def pytest_report_header(config):
+    """Report testsuite metadata"""
+    header = [""]
+    for key in sorted(config._metadata.keys()):
+        header.append(f"testsuite: {key}={config._metadata[key]}")
+        _junit_testsuite_property(config, key, config._metadata[key])
+    for key in sorted([k for k in os.environ if k.startswith("_3SCALE_TESTS")]):
```
pehala, line 194: unnecessary list

testsuite/tests/conftest.py
```diff
@@ -122,104 +122,78 @@ def pytest_collection_modifyitems(session, config, items):
...
-        f"testsuite: environment = {environment}",
-        f"testsuite: openshift = {openshift}",
-        f"testsuite: project = {project}",
-        f"testsuite: threescale = {threescale}",
-        f"testsuite: for 3scale version = {version}",
-        ]
+    title = _settings["reporting"]["title"] % f"{title} {namespace} {version}"
+
+    metadata.update({
+        "env_for_dynaconf": settings["env_for_dynaconf"],
```
pehala, line 150: `_settings`?

testsuite/tests/conftest.py
mganisin: ... what breaks ... nothing

testsuite/tests/conftest.py
mganisin: `env_for_dynaconf` is always present

mdujava (approved):

pehala (approved):

PR #336, opened 2022-07-24
Introduce custom markers to skip tests based on version

mganisin:
Try to use custom markers and special handling to skip tests that aren't
relevant to specific 3scale version.

Builtin skipif mark is better as there is no new code and no new
behavior, however... Having two imports (that must be ignored in checks)
and repeating still same conditions is pretty inconvenient. Furthermore
custom marks create standardized behavior that doesn't allow "creative"
conditions.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -8,6 +8,10 @@ markers =
 	required_capabilities(capability1, capability2, ...): List of capabilities that are required for running this test
 	issue: Reference to covered issue
 	nopersistence: Tests incompatible with persistence plugin
+	require_version: Run on this or later version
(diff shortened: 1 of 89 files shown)


mganisin:
a thought to record... Question is whether `until_version` & `after_version` would be more handy

mganisin:
Closing for now, this was just a brainstorming kickoff how these markers should look like. `(require|before)_version` looks right approach (and due to easier use, code reduction and elimination of exceptions also better than builtin `skipif`). However the effort to rewrite all (and avoid merge conflicts) or introduce different style for newer tests isn't worth at this moment

PR #352, opened 2022-07-29
Add test for product search

JaurbanRH:


testsuite/tests/ui/search/test_account_search.py
```diff
@@ -0,0 +1,95 @@
+"""
+Rewrite of spec/ui_specs/users/users_search_spec.rb
```
dhlavac, line 2: Imho leftover from user search test

jsmolar:
Quick question @dhlavac. Do we need to test the search for accounts, products, and backends? They all use sphinx. I assume it would be enough to test just one of them.

dhlavac:
@jsmolar we need at least tests for Acounts and (Products or/and Backends) as search works differently on this pages

jsmolar (approved):

dhlavac (changes requested):
You have not removed the old account search test so we have now 2 account search tests in test suite(one in search ui/dir and one in ui dir), I highly recommend separating tests to separate PR as you have done before.

testsuite/tests/ui/search/test_account_search.py
```diff
@@ -94,3 +93,24 @@ def test_search_non_existing_value(request, login, navigator, custom_account):
...
+@pytest.mark.xfail
+@pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-8468")
+def test_search_short_keyword(login, navigator, custom_account, request):
+    """
+    Preparation:
+        - Create custom account with short keyword (less than 3 characters)
+    Test if:
+        - you search account by specific org_name it will return the correct one
+    """
+    org_name = blame(request, "org-1")
```
dhlavac, line 108: Better to use at least 2 characters due to the limitations of Sphinx e.g. "org-cz"

testsuite/tests/ui/search/test_account_search.py
```diff
@@ -94,3 +93,24 @@ def test_search_non_existing_value(request, login, navigator, custom_account):
...
+    params = dict(org_name=org_name, username="username", email="email@anything.invalid",
```
dhlavac, line 109: Please change the username to something else than "username" as it is already used in this test module.

testsuite/tests/ui/search/test_product_search.py
```diff
@@ -0,0 +1,70 @@
...
+@pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-8562")
+def test_search_product(login, navigator, request, custom_ui_product):
+    """
+    Preparation:
+        - Create custom account
+        - Create custom application
+    Test if:
+        - you search account by org_name it will return the correct one
+        - you search account by username it will return the correct one
+        - you search account by email it will return the correct one
```
dhlavac, line 21: Product are no searchable by email or application please remove incorrect docstring description

dhlavac (changes requested):

testsuite/tests/ui/search/test_product_search.py
```diff
@@ -0,0 +1,70 @@
...
+    products.search(f"{name}_name2")
+    assert products.table.row()[0].text == f"{name}_name2"
+
+
+def test_search_non_existing_value(request, login, navigator, custom_service):
+    """
+    Preparation:
+        - Create custom service
+    Test if:
+        - you search account by non-existing-value it won't return anything
```
dhlavac, line 63: Also incorrect description

testsuite/ui/views/admin/product/__init__.py
```diff
@@ -14,6 +14,13 @@ class ProductsView(BaseAdminView):
     table = PatternflyTable("//*[@id='products']/section/table", column_widgets={
         "Name": Text("./a")
     })
+    search_bar = TextInput(locator="//input[@type='search']")
+    search_button = GenericLocatorWidget("//button[contains(@aria-label,'search')]")
+
+    def search(self, value: str):
+        """Search in account table by given value"""
```
dhlavac, line 21: Search in product table

dhlavac (changes requested):

dhlavac (approved):

testsuite/tests/ui/search/test_account_search.py
```diff
@@ -58,20 +59,22 @@ def test_search_multiple_accounts(login, navigator, custom_account, request):
         - you search account by specific name it will return the correct one
     """
     name = blame(request, "")
-    params = [dict(org_name=f"{name}_name", username="username", email="email@anything.invalid", password="123456"),
-              dict(org_name=f"{name}_name2", username="username2", email="email2@anything.invalid", password="123456"),
-              dict(org_name=f"{name}", username="user", email="mail@anything.valid", password="123456"),
-              dict(org_name=blame(request, "RedHat"), username="random", email="random@random.random",
-                   password="123456")
+    params = [dict(org_name=f"{name}_name", username=blame(request, "username"),
```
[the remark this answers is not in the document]
JaurbanRH: I don't think that this is necessary change and it would only complicate more things. @dhlavac @pehala opinion?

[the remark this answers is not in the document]
jsmolar: By doing this ^^ you don't repeat the same exact code 4 times and it is obvious what names are your accounts using, thus reading this and asserts are clear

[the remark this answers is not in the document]
mganisin: Personally I think all irrelevant unrelated changes should be remove from this PR at first place and only "test for product search" code should be included. After that a discussion (and decision) should be little bit easier.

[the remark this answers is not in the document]
dhlavac: @jsmolar I agree with you, it would be easier to read that test, and absolutely agree with @mganisin !!! It would be great to move unrelated changes to separate PR! (as I mentioned several times :) )

PR #364, opened 2022-08-09
pytest html for ui (all actually)

mganisin:
- Instakk also pytest-html
- junitxml name based just on basename
- Add UI screenshot reference to pytest-html report
- Automatically generate html report for UI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,7 @@ ENV REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
 ENV SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt
 ENV WORKON_HOME=/opt/workdir/virtualenvs
 ENV junit=yes
+ENV html=yes
 ENV resultsdir=/test-run-results
 
 RUN echo umask 002 >>$HOME/.bashrc \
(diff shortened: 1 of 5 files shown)


dhlavac:
What about default gitignore result dir in test suite, when there is none specified

mdujava (dismissed):

jsmolar (dismissed):

dhlavac (approved):

dhlavac (changes requested):
When running tests from a specific directory(not the testsuite root) no screenshots are added to the HTML report. Could it be possible, to save the report by default to the same directory as screenshots are saved?

mganisin:
> When running tests from a specific directory(not the testsuite root) no screenshots are added to the HTML report. Could it be possible, to save the report by default to the same directory as screenshots are saved?

@dhlavac path to report is set on commandline, unlikely good idea to alter it as it wouldn't have to be clear when to modify it and when not. It's better idea to set it on command line with absolute path.

The issue with links to the screenshots should be fixed, they should be present correctly in html report now. Try it, please

mdujava (approved):

dhlavac (approved):

dhlavac:
LGTM!!!

PR #372, opened 2022-08-23
Changes in account search test

JaurbanRH:


testsuite/tests/ui/search/test_account_search.py
```diff
@@ -58,20 +73,18 @@ def test_search_multiple_accounts(login, navigator, custom_account, request):
         - you search account by specific name it will return the correct one
     """
     name = blame(request, "")
```
dhlavac, line 75: Aren't you missing the first string or are all names supposed to start with a dash?

testsuite/tests/ui/search/test_account_search.py
```diff
@@ -58,20 +73,18 @@ def test_search_multiple_accounts(login, navigator, custom_account, request):
...
-              dict(org_name=f"{name}", username="user", email="mail@anything.valid", password="123456"),
-              dict(org_name=blame(request, "RedHat"), username="random", email="random@random.random",
-                   password="123456")
-              ]
-    for param in params:
-        custom_account(param)
+    custom_account(name)
+    custom_account(f"{name}_name1")
+    custom_account(f"{name}_name2")
+    custom_account("RedHat")
```
dhlavac, line 79: Imho some name which does not contain RedHat should be better because when you run the test with RHSSO tests(or manual using a RedHat email address) they by default create the organization RedHat. So if tests for RHSSO fail, resources won't be cleaned and there may be unexpected results.

dhlavac (changes requested):

dhlavac (approved):

mganisin (approved):

PR #386, opened 2022-09-06
Fix RHSSO Performance tests

pehala:
To sum it up, Httpbin-go knows only `/anyhing/` path and not `/anything` (expects a trailing slash), but on `/anything` it returns 301 with relative redirect to `/anything/` -> which our client uses and asks APIcast again with the new path (`APICAST/anything`) -> this doesn't conform with any mapping rules and returns 404....


dhlavac (approved):
LGTM!

JaurbanRH (approved):

PR #398, opened 2022-09-12
Unify gathering of echo-api name in maintenance_mode test

mganisin:
Don't hardcode any piece of hostname as the backend service can change
even is symbolic name is 'echo_api'

### DIFF
--- a/testsuite/tests/apicast/policy/maintenance_mode/test_maintenance_mode_policy_host.py
+++ b/testsuite/tests/apicast/policy/maintenance_mode/test_maintenance_mode_policy_host.py
@@ -5,6 +5,7 @@ Expected: to return specified code eg 328 and message of service unavailability.
 """
 
 from typing import Tuple
+from urllib.parse import urlparse
 
 import pytest
 import pytest_cases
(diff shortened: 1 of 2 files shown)


JaurbanRH (approved):

testsuite/tests/apicast/policy/maintenance_mode/test_maintenance_mode_policy_host.py
```diff
@@ -20,16 +21,22 @@
     pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-6552")]
 
 
+@pytest.fixture
+def echo_api_base_url(private_base_url):
+    """Return URL of echo_api"""
+    return private_base_url("echo_api")
```
pehala, line 27: Since you are already changing this, wouldn't it make sense to changing to the actual backend and not an alias?

mganisin: No, this should be working also in evironments where echo_api is echo-api (maybe saas)

pehala (approved):

PR #401, opened 2022-09-12
test http proxy

mganisin:
- Use explicit service accepting http
- Use router based URL instead of .svc

### DIFF
--- a/testsuite/tests/apicast/parameters/http_proxy/test_http_proxy.py
+++ b/testsuite/tests/apicast/parameters/http_proxy/test_http_proxy.py
@@ -22,7 +22,7 @@ def gateway_kind():
 @pytest.fixture(scope="module")
 def service_proxy_settings(private_base_url):
     "Dict of proxy settings to be used when service created"
-    return rawobj.Proxy(private_base_url("httpbin_service"))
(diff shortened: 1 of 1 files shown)


JaurbanRH (approved):

JaurbanRH (approved):

testsuite/tests/apicast/parameters/http_proxy/test_http_proxy.py
```diff
@@ -22,7 +22,7 @@ def gateway_kind():
 @pytest.fixture(scope="module")
 def service_proxy_settings(private_base_url):
     "Dict of proxy settings to be used when service created"
-    return rawobj.Proxy(private_base_url("httpbin_service"))
+    return rawobj.Proxy(private_base_url("go-httpbin"))
```
dhlavac, line 25: shouldnt be there `"echo-api+svc:9292"`

mganisin: No it should not. a) This is part of effort to eliminate dependency on echo-api. b) this PR is screwed little bit, please look at commit message directly for explanation

dhlavac (approved):

PR #407, opened 2022-09-15
Use unique name for a fixture

mganisin:
private_base-url is a fixture used in global conftest as well, it's
behavior can't change. Unique name (different fixture) must be used for
local modification


testsuite/tests/apicast/policy/proxy/test_fuse_proxy_policy.py
```diff
@@ -23,23 +23,23 @@ def policy_settings(testconfig):
 
 
 @pytest.fixture(scope="module")
-def private_base_url(private_base_url):
+def the_private_base_url(private_base_url):
```
pehala, line 26: How about `backend_url`?

dhlavac (approved):

pehala (approved):

PR #418, opened 2022-09-21
Refactor of SSL verify for webdriver

jsmolar:


JaurbanRH (approved):

testsuite/ui/webdriver.py
```diff
@@ -15,15 +15,13 @@
...
+    def __init__(self, source, accept_insecure_certs, remote_url=None, binary_path=None):
         self.source = source
         self.remote_url = remote_url
         self.binary_path = binary_path
         self.webdriver = None
         self.options = ChromeOptions()
-
-        if not ssl_verify:
-            self.options.set_capability("acceptInsecureCerts", True)
+        self.options.set_capability("acceptInsecureCerts", accept_insecure_certs)
```
pehala, line 24: Why did you change the name and not simply used `self.options.set_capability("acceptInsecureCerts", not ssl_verify)`?

jsmolar: Because I would need to do this for both ff and chrome. By doing this, we have only one place where to define the correct value (note that, bot chrome and ff are using this the same way) and it is easily readable for the user, then to think what happens with negation.
With line `accept_insecure_certs = not ssl_verify` in the parent class this negation is more obvious IMHO.

dhlavac (approved):

PR #426, opened 2022-10-06
Use echo_api alias

mganisin:
echo-api != echo_api. 'echo_api' is an alias that is changed to a
service providing features of echo-api. And 'echo-api' is a request
specifically for echo-api.


JaurbanRH (approved):

azgabur (approved):
I approve, ~but the test_wizard test is failing on `page_title.text` string comparation.
`AssertionError: assert 'Good. Next, make a test GET request' == 'Congratulations, you are running on 3scale :-)'`
Maybe can be fixed in other PR?~ There is no problem in new dev build.

dhlavac (approved):

PR #429, opened 2022-10-07
Fixed adding user key to oas3 and removed skip validation for oas3 tests

jsmolar:


dhlavac:
Why do we need this change? IMHO it is just an equivalent form of notation.

jsmolar:
@dhlavac Yes, we need this. With this change, oas3 is valid and you don't need to skip validation when creating a new spec in 3scale

dhlavac (approved):

JaurbanRH (approved):

PR #433, opened 2022-10-19
Fix credentials logging

dhlavac:
Temporarily fix before https://github.com/RedHatQE/widgetastic.core/pull/227 will be merged.

### DIFF
--- a/testsuite/ui/views/common/login.py
+++ b/testsuite/ui/views/common/login.py
@@ -14,7 +14,11 @@ class LoginForm(View):
     password_label = Text('//input[@id="session_password"]/preceding-sibling::label')
     submit = ThreescaleSubmitButton()
 
-    def do_login(self, name, password):
+    def fill_passwd(self, passwd: str):
+        """Method fill password field with password logging disabled"""
+        self.password_field.fill(passwd, sensitive=True)
+
+    def do_login(self, name: str, password: str):
         """
         Method handle login to 3scale admin portal
         :param name: User username for login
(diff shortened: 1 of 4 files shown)


JaurbanRH (approved):

testsuite/ui/views/common/login.py
```diff
@@ -22,7 +31,9 @@ def do_login(self, name, password):
         :return DashboardView page object
         """
         self.username_field.fill(name)
+        LOGGER.disabled = True
         self.password_field.fill(password)
```
jsmolar, line 35: use `self.fill_passwd` method here

pytest.ini
```diff
@@ -21,5 +21,6 @@ junit_logging=all
 junit_family=xunit2
 junit_log_passing_tests=0
 asyncio_mode=auto
+log_cli = true
```
jsmolar, line 24: why do we need `log_cli`?

dhlavac: Not necessary but logs with it provide much more info for investigation, at least for UI tests. @mganisin what do you think?

mganisin: good for your local PYTEST_ADDOPTS, not that great as global default. Even if considered good, certainly should not be introduced here, but in separate MR.

jsmolar (dismissed):

mganisin:
@dhlavac this requires your care

dhlavac:
blocked by https://issues.redhat.com/browse/IQE-2070

JaurbanRH (approved):

jsmolar (approved):

PR #437, opened 2022-10-26
Added test for Application plan change

dhlavac:


JaurbanRH (dismissed):

testsuite/tests/ui/applications/test_application_plans.py
```diff
@@ -0,0 +1,51 @@
...
+    assert plan
+    assert plan['name'] == name
+
+
+# pylint: disable=unused-argument
+@pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-825")
+@pytest.mark.skipif("TESTED_VERSION < Version('2.13')")
+def test_default_app_plan_change(login, service, custom_app_plan, navigator):
+    """
+    Test cover functionality of  feature 825 - default application plan change
```
jsmolar, line 29: We don't need to mention the issue here. that is what the mark issue is for.
I'd rather see short description.

testsuite/tests/ui/applications/test_application_plans.py
```diff
@@ -0,0 +1,51 @@
...
+    app_plans_view = navigator.navigate(ApplicationPlansView, product=service)
+
+    assert not app_plans_view.change_plan_button.is_enabled
+    app_plans_view.default_plan_select.item_select(plan1["name"])
+    assert app_plans_view.change_plan_button.is_enabled
+    app_plans_view.change_plan_button.click()
+    assert app_plans_view.notification.is_displayed
+    assert app_plans_view.notification.string_in_flash_message("the default plan has been changed.")
+
+    assert not app_plans_view.change_plan_button.is_enabled
```
jsmolar, line 47: What about some API check or real request to verify that plan was changed in UI and in the system as well

testsuite/tests/ui/applications/test_application_plans.py
dhlavac: Yes, we need it. It is a feature request and needs to be tracked.

testsuite/tests/ui/applications/test_application_plans.py
dhlavac: Currently out of scope of this test

testsuite/ui/views/admin/product/application.py
```diff
@@ -15,6 +16,20 @@ class ApplicationPlansView(BaseProductView):
         "Name": Text("./a")
     })
     create_button = ThreescaleCreateButton()
+    change_plan_button = ThreescaleSubmitButton()
+    default_plan_select = Select()
+    notification = View.nested(FlashMessage)
+
+    def change_default_plan(self, plan_name: str):
```
JaurbanRH, line 23: Why you add this function? You are not using it anywhere in the test.

dhlavac: I have planned to add an extended test, that changes the default plan but after discussion, it will be part of the next PR. So this code only enables the basic functionality of changing the default plan.

JaurbanRH: So it's not in scope of this PR and should be part of the next PR right?

dhlavac: Removed

JaurbanRH (approved):

jsmolar (approved):

PR #438, opened 2022-10-27
New wasm image and WasmPlugin custom resource support

azgabur:
This PR is dependent on yet unreleased wasm image. 
(for testing purposes, @pehala 's image can be used `quay.io/phala/threescale-wasm-auth:v16`)

- Adds support for new WasmPlugin
- Fixes gateways resource not deleting
- Removes mapping rules synchronization, as its not needed in new wasm image 
- Removes backend token synchronization, as its also not needed
- Remove special status code handling in tests as now correct status codes are returned
- Reduce openshift calls in credentials synchronization and rollout only if they were changed
- Add image pull secret support for wasm (_currently untested_)

### DIFF
--- a/config/settings.yaml
+++ b/config/settings.yaml
@@ -13,6 +13,7 @@ default:
         template: apicast.yml
       WASMGateway:
         image: "registry.redhat.io/openshift-service-mesh/3scale-auth-wasm-rhel8:0.0.1"
+        pull_secret: "pull-secret"  # name of pull secret resource in same namespace as httpbin
       default:
         kind: "SystemApicast"
   rhsso:
--- a/testsuite/gateways/wasm/__init__.py
+++ b/testsuite/gateways/wasm/__init__.py
@@ -96,11 +96,10 @@ class WASMGateway(AbstractGateway):
     # pylint: disable=unused-argument, too-many-arguments
     def _create_api_client(self, application, endpoint, verify, cert=None, disable_retry_status_list=None):
         ext = self.extensions[application.service["id"]]
-        ext.synchronise_mapping_rules()
         ext.synchronise_credentials()
 
         # wait needed for WASM to load new configuration, fixme: should be replaced with wait for wasm sync
-        ext.httpbin.deployment(f"dc/{ext.httpbin_name}").rollout()
+        ext.httpbin.deployment(f"dc/{ext.httpbin_name}").wait_for()
         return ServiceMeshHttpClient(app=application,
                                      cert=cert,
                                      disable_retry_status_list=disable_retry_status_list,
--- a/testsuite/gateways/wasm/__init__.py
+++ b/testsuite/gateways/wasm/__init__.py
@@ -96,10 +96,9 @@ class WASMGateway(AbstractGateway):
     # pylint: disable=unused-argument, too-many-arguments
     def _create_api_client(self, application, endpoint, verify, cert=None, disable_retry_status_list=None):
         ext = self.extensions[application.service["id"]]
-        ext.synchronise_credentials()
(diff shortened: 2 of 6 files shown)


testsuite/gateways/wasm/__init__.py
```diff
@@ -96,11 +96,10 @@ def remove_policy(self, name: str):
     # pylint: disable=unused-argument, too-many-arguments
     def _create_api_client(self, application, endpoint, verify, cert=None, disable_retry_status_list=None):
         ext = self.extensions[application.service["id"]]
-        ext.synchronise_mapping_rules()
-        ext.synchronise_credentials()
+        if ext.synchronise_credentials():
+            ext.httpbin.deployment(f"dc/{ext.httpbin_name}").rollout()
+            ext.httpbin.deployment(f"dc/{ext.httpbin_name}").wait_for()
```
pehala, line 101: I believe rollout also waits for the resource

testsuite/gateways/wasm/extension.py
```diff
@@ -106,14 +104,27 @@ def synchronise_mapping_rules(self):
...
+    def replace_credentials(self, credentials):
+        """Replaces previous credentials and deploys new into the extension.
+            `credentials` contains list of credentials with attributes:
             label: one of "user_key" or "app_id" or "app_key",
-            credential_location: one of "query_string", "header" or "authorization"
+            credential_location: one of "query_string", "header", "authorization", or "oidc"
             key: search key
         """
         ops = []
+        for label in ["user_key", "app_id", "app_key"]:
```
pehala, line 115: [Minor] Use set

testsuite/resources/service_mesh/plugin.yaml
```diff
@@ -0,0 +1,89 @@
...
+objects:
+- apiVersion: extensions.istio.io/v1alpha1
+  kind: WasmPlugin
+  metadata:
+    name: ${NAME}
+    labels:
+      app: ${LABEL}
+  spec:
+    url: oci://${IMAGE}
+    phase: AUTHZ
```
pehala, line 14: If used with the released image, we will also need to add a field containing a pullSecret.

testsuite/gateways/wasm/__init__.py
azgabur: If so, I removed the wait

testsuite/gateways/wasm/extension.py
azgabur: Addressed

azgabur: It doesnt really make a difference and set usage generates lint warning 
` C0208: Use a sequence type when iterating over values (use-sequence-for-iteration)`
Reverting

testsuite/resources/service_mesh/plugin.yaml
azgabur: Added support for this

pehala (approved):

config/settings.yaml
```diff
@@ -13,6 +13,7 @@ default:
         template: apicast.yml
       WASMGateway:
         image: "registry.redhat.io/openshift-service-mesh/3scale-auth-wasm-rhel8:0.0.1"
+        pull_secret: "pull-secret"  # name of pull secret resource in same namespace as httpbin
```
dhlavac, line 16: Please add also WASMGate configuration to settings tlp.

pehala: I agree, however I would add it in a separate PR

azgabur: Opened new PR #461 for this

dhlavac (approved):

PR #442, opened 2022-10-30
Little help with encrypted config/.secrets.yaml

mganisin:


pehala (approved):

Makefile
```diff
@@ -234,6 +235,16 @@ tools: export SHARED_NAMESPACE ?= tools
 tools:
 	./ext/testsuite-tools/run.sh
 
+define n
+
+
+endef
+check-secrets.yaml:
+ifeq ($(shell grep -lIL . config/.secrets.yaml), config/.secrets.yaml)
```
dhlavac, line 243: Maybe also check for the presence of .secrets.yaml?

dhlavac (changes requested):

Makefile
mganisin: That should not be encrypted, should be? Maybe you think about `cp config/.secrets.yaml .secrets.yaml` but that is totally unexpected when README and also this error message say "delete it". I am afraid it's far beyond scope, it could also check couple of env variables as well as `.env` file (and that in particular can be quite dangerous in topdir)

dhlavac (approved):

PR #448, opened 2022-11-02
Ensure correct order of fixtures initialization to prevent cleanup failures

mganisin:


mganisin:
@mkudlej @mdujava I think I remember src threescale instance was configured separately, therefore custom_tenant was wasn'ŧ required directly to avoid dependency on openshift. But now src threescale seems to be standard instance used in the testsuite therefore  that indirect dependency is not needed anymore and this PR is safe. Correct?

pehala (approved):

mdujava (approved):
This is fine for me. But this change will introduce that you need to have oc set up for tests in a toolbox subtree.

PR #452, opened 2022-11-04
Add web browser version to metadata

mganisin:
Adding information to metadata in ui conftest won't print this info in
the header on the console output, but it will be included in html report
(if generated) as well as in xml (if generated), so the information will
be retained and available.

### DIFF
--- a/testsuite/tests/ui/conftest.py
+++ b/testsuite/tests/ui/conftest.py
@@ -45,7 +45,7 @@ def webdriver():
 
 
 @pytest.fixture(scope="module")
-def browser(webdriver, request):
+def browser(webdriver, request, metadata):
     """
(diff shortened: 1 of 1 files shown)


JaurbanRH (approved):

testsuite/tests/ui/conftest.py
```diff
@@ -55,6 +55,8 @@ def browser(webdriver, request):
     """
     browser = ThreeScaleBrowser(selenium=webdriver.start_session())
     request.addfinalizer(webdriver.finalize)
+    caps = webdriver.webdriver.webdriver.caps
```
jsmolar, line 58: this looks like shit, but what can we do :D

jsmolar (approved):

PR #474, opened 2022-12-14
Tenant impersonation test added

dhlavac:


testsuite/tests/ui/master/test_tenant.py
```diff
@@ -119,3 +119,25 @@ def test_resume_tenant(master_login, navigator, master_threescale, custom_tenant
...
+    Test check impersonation functionality:
+        - create tenant
+        - login into master portal
+        - impersonate new created tenant
+        - assert that the impersonation was successful
+            - dashboard is displayed
+            - url is correct
+    """
+    tenant = custom_tenant()
+    tenant.wait_tenant_ready()
```
JaurbanRH, line 135: This functionality can be moved to separate fixture as we do for other `custom_entities`. Also `wait_tenant_ready()` is already called in the `custom_tenant()` function.

jsmolar (approved):

mganisin (approved):

PR #475, opened 2022-12-21
Wasm authorities tests

azgabur:
This PR:

- Add authorities modification functions for WASM
- New fixture in conftest for default extension associated with default service
- Fix bug with missing else branch in ingress_url property
- Add ingress_alias_url property
- Add tests for authorities in WASM

### DIFF
--- a/testsuite/gateways/wasm/extension.py
+++ b/testsuite/gateways/wasm/extension.py
@@ -19,6 +19,7 @@ class WASMExtension:
         self.backend_endpoint = backend_endpoint
         self.base_path = resources.files('testsuite.resources.service_mesh')
         self._ingress_url = None
+        self._ingress_alias_url = None
 
         self.service = service
         self.label = f"{parent_label}-{self.service['id']}"
@@ -51,9 +52,22 @@ class WASMExtension:
             route = self.mesh.routes["istio-ingressgateway"]
             if "tls" in route["spec"]:
                 self._ingress_url = "https://" + route["spec"]["host"]
-
-            self._ingress_url = "http://" + route["spec"]["host"]
+            else:
+                self._ingress_url = "http://" + route["spec"]["host"]
         return self._ingress_url
+
+    @property
+    def ingress_alias_url(self):
+        """Returns alias url dynamically made by service mesh."""
+        if not self._ingress_alias_url:
+            for route in self.mesh.routes:
+                if self.httpbin_name in route["spec"]["host"]:
+                    alias_route = route
+            if "tls" in alias_route["spec"]:
+                self._ingress_alias_url = "https://" + alias_route["spec"]["host"]
+            else:
+                self._ingress_alias_url = "http://" + alias_route["spec"]["host"]
+        return self._ingress_alias_url
 
     def add_mapping_rules(self, rules):
         """Adds mapping rule into the extension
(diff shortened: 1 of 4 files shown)


testsuite/gateways/wasm/extension.py
```diff
@@ -212,6 +226,53 @@ def synchronise_credentials(self):
             return True
         return False
 
+    def add_authorities(self, authorities, wait=True):
```
pehala, line 229: Couldn't it be fully replaced by `replace_authorities`?

testsuite/tests/service_mesh/auth/conftest.py
```diff
@@ -11,3 +11,9 @@ def no_auth_status_code(staging_gateway):
     if isinstance(staging_gateway, WASMGateway):
         return 403
     return 401
+
+
+@pytest.fixture(scope="module")
+def extension(staging_gateway, service) -> WASMExtension:
```
pehala, line 17: I would move it to the lowest possible conftest/test in this case, `test_authorities`

testsuite/tests/service_mesh/auth/wasm/test_authorities.py
```diff
@@ -0,0 +1,79 @@
...
+def client_alias(api_client, extension) -> ServiceMeshHttpClient:
+    """Returns api client using first alias url as ingress to service mesh."""
+    client: ServiceMeshHttpClient = api_client()
+    client.root_url = extension.ingress_alias_url
+    return client
+
+
+def test_alias(extension, client_ingress, client_alias):
+    """This test sets extension authorities to only accept alias host."""
+    extension.replace_authorities([urlsplit(extension.ingress_alias_url).netloc])
```
pehala, line 30: This seems dangerous as the tests cannot be executed in parallel (Should not cause problems for now as we run each module together). How hard would it be to create an extension per test?

testsuite/gateways/wasm/extension.py
```diff
@@ -51,10 +52,23 @@ def ingress_url(self):
             route = self.mesh.routes["istio-ingressgateway"]
             if "tls" in route["spec"]:
                 self._ingress_url = "https://" + route["spec"]["host"]
```
pehala, line 54: `return "https://" + route["spec"]["host"]`

testsuite/gateways/wasm/extension.py
```diff
@@ -51,10 +52,23 @@ def ingress_url(self):
...
-
-            self._ingress_url = "http://" + route["spec"]["host"]
```
pehala, line 55: `return "http://" + route["spec"]["host"]`

testsuite/gateways/wasm/extension.py
azgabur: Could be. I parametrized the `replace_authorities` function

testsuite/gateways/wasm/extension.py
azgabur: This would cause the variable `self._ingress_url` to be useless as there is no assignment to it.

testsuite/gateways/wasm/extension.py
azgabur: ditto

testsuite/tests/service_mesh/auth/wasm/test_authorities.py
azgabur: Would be just bit more complicated. 
By creating `service` per test there will be `extension` per test. But than every test should also have its own `client_alias` as alias url is unique per extension. I will rewrite it that way than.

testsuite/tests/service_mesh/auth/conftest.py
azgabur: I removed the function as is no longer needed.

testsuite/tests/service_mesh/auth/wasm/test_authorities.py
azgabur: I did rewrite the test the way I described above. It has drawback of longer test run because of additional service/extension creations and rollouts.

testsuite/tests/service_mesh/auth/wasm/test_authorities.py
```diff
@@ -4,76 +4,106 @@
...
 
 
 @pytest.fixture(scope="module")
-def client_ingress(api_client, extension) -> ServiceMeshHttpClient:
-    """Returns api client using default ingress_url as ingress to service mesh."""
-    client: ServiceMeshHttpClient = api_client()
-    client.root_url = extension.ingress_url  # redundant, as its default, but included for better readability
-    return client
+# pylint: disable=too-many-arguments
+def custom_extension_and_app(custom_service, custom_app_plan, custom_application,
```
pehala, line 15: The same result should be achievable with only rewriting existing fixtures to be `function` scope

azgabur: Addressed

pehala (approved):

mganisin (approved):

PR #477, opened 2023-01-02
mark some special ui tests as sandbag

mganisin:


dhlavac:
@mganisin I think also service discovery test can be flagged as sandbag

jsmolar:
> @mganisin I think also service discovery test can be flagged as sandbag

Agree, with the description that it does not work on RHOAM

dhlavac (approved):

mdujava (approved):

PR #480, opened 2023-01-16
Option to run all the tests in dedicated tenant

mganisin:
By default the tests are executed on default tenant or on that provided
by config.

Since now there is the option to run the tests on new/fresh tenant that
is created at the beginning of run and deleted at the end.

This covers the cases when custom tenant creation is covered by all the
tests or if it is easier to get master token than admin token (typically
true on deployments where a tenant for testing purposes is not present).

A caveat is that the testsuite creates as many tenant as many "threads"
are used to run it ('-n' option of xdist).

### DIFF
--- a/config/settings.yaml.tpl
+++ b/config/settings.yaml.tpl
@@ -53,6 +53,8 @@ default:
       polarion_project_id: PROJECTID
       polarion_response_myteamsname: teamname
   fixtures:
+    threescale:
+      private-tenant: False  # if true standalone tenant is created to run all the tests
     jaeger:
       url: "" # route to the jaeger-query service for the querying of traces
       config:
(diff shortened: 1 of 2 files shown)


mkudlej:
@mganisin What are the reasons to create new tenants directly in testsuite and not to create this really complicated 3scale object outside of testsuite(and not to use configuration options for first tenant as you've mentioned)?

mganisin:
> This covers the cases when custom tenant creation is covered by all the
> tests or if it is easier to get master token than admin token (typically
> true on deployments where a tenant for testing purposes is not present).

This covers the cases when custom tenant creation is covered by all the
tests or if it is easier to get master token than admin token (typically
true on deployments where a tenant for testing purposes is not present).

JaurbanRH (approved):

mdujava (approved):

pehala (commented):
It seems to work, however, there is no indication anywhere that the option was used e.g. the testsuite header will still list the same admin URL.

PR #500, opened 2023-02-08
Add backoff to select_by_value of dropdown widget

mganisin:
Some test(s) fail on this call with excetion when such item does not
exist. Might be because of quick navigation (or slow rendering). In some
environments happens regularly, in others never. Retry (as done in other
widget valls) should/could help.


JaurbanRH (approved):

mganisin:
won't work

PR #509, opened 2023-02-16
don't set shell explicitly in github_pre_release

mganisin:


.github/workflows/github_pre_release.yaml
```diff
@@ -13,7 +13,6 @@ jobs:
         with:
           fetch-depth: 0
       - name: github_pre_release
-        shell: bash
```
mdujava, line 16: I think you need `bash {0}`, isn't bash default value, which will add the pipefail option?

PR #512, opened 2023-02-21
Braintree cc setup

jsmolar:
Improving Braintree gateway billing tests. 

1. By small refactoring this PR simplifies asserts used by billing tests.
2. This Pr also introduced an overall better structure needed for further test expansion.
3. Adding tests that are focused on adding cards. With an emphasis on adding a new card for an account. In the future, we would like to add tests for payment verification and occurring payments.


testsuite/ui/views/devel/settings/braintree.py
```diff
@@ -59,18 +62,35 @@ def _iframe_fill(self, frame_id, widget, value):
...
+        self.cancel_btn.click()
+        self.browser.selenium.switch_to.default_content()
+        self._wait_challenge_completed()
+
+    def _wait_challenge_completed(self):
+        """Wait for `BraintreeCCView` to reflect all changes when 3DS challenge was completed (or closed)"""
+        self.browser.wait_for_element(
+            ".//dt[normalize-space(.)='Credit card number']|"
+            ".//p[@class='alert alert-danger']",
+            timeout=20
```
josemigallas, line 92: The dialog's default timeout is 60 seconds (see https://cardinaldocs.atlassian.net/wiki/spaces/CC/pages/1409568/Configurations#Configurations-RootLevelConfiguration). One of the testing credit cards is used in this scenario so I would at least pass `timeout` as an argument.

testsuite/tests/ui/devel/billing/braintree/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    return _braintree
+
+
+@pytest.fixture(scope="module")
+def setup_card(account, custom_devel_login, billing_address, navigator):
+    """Add credit card and billing address for the user in Developer portal"""
+    def _setup(cc_number, verify_3ds=False):
+        custom_devel_login(account=account)
+        cc_view = navigator.navigate(BraintreeCCView)
+        cc_view.add_cc_details(billing_address, CreditCard(cc_number, 123, 1, 24))
```
josemigallas, line 44: According to Braintree docs, the expiration date must be current year +3. From https://developer.paypal.com/braintree/docs/guides/3d-secure/migration/javascript/v3#client-side-sandbox-testing:
> IMPORTANT
Use the card numbers below to test 3DS2 on your client side in sandbox. Any other test cards will result in 3DS verification results that are not in the 3DS2 format.
>
>For expiration year values for all test cards in the table below, use the current year +3. The expiration month should always be 01. For example, if the current year is 2023, a valid test value for the expiration date would be 01/26.

testsuite/tests/ui/devel/billing/braintree/test_3ds_challenge.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    account.credit_card_delete()
+    cc_view = setup_card("4000000000001109", True)
+    assert cc_view.alert() == "An error occurred, please review your CC details or try later."
+
+    invoice = create_api_invoice()
+    with pytest.raises(ApiClientError) as exc_info:
+        invoice.charge()
+
+    assert exc_info.type is ApiClientError
```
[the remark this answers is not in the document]
pehala: Figure it out :D

testsuite/ui/views/devel/settings/braintree.py
jsmolar: Great point, I will add this scenario in the next PR (this one is already large)

JaurbanRH (approved):

dhlavac (changes requested):
I agree with @pehala on the named parameter in `braintree_gateway` fixture. Also maybe an enum for cards would be great to recognize card purpose easier

jsmolar:
> I agree with @pehala on the named parameter in `braintree_gateway` fixture. Also maybe an enum for cards would be great to recognize card purpose easier
Agree with the fixture.
CC's purpose is apparent from the test doc itself, but I am open to any idea.

testsuite/tests/ui/devel/billing/braintree/conftest.py
mganisin: @jsmolar `current_year + 3` looks like pretty easy fix :-)

dhlavac (approved):

pehala (dismissed):

testsuite/billing.py
```diff
@@ -57,10 +57,10 @@ def merchant_currency(self):
         merchant_accounts = list(self.gateway.merchant_account.all().merchant_accounts.items)
         return [x for x in merchant_accounts if x.default is True][0].currency_iso_code
 
-    def assert_payment(self, acc_invoices):
+    def assert_payment(self, invoice):
         """Compare 3scale invoice and Braintree transaction"""
-        invoice = acc_invoices[0]
         transaction = self.gateway.transaction.search(braintree.TransactionSearch.order_id == str(invoice['id']))
+        self.gateway.customer.search(braintree.CustomerSearch.id.contains("342"))
```
pehala, line 63: why?

pehala (approved):

mganisin (approved):

PR #518, opened 2023-03-01
Add test for wrong realm

pehala:
* Addresses [THREESCALE-9009](https://issues.redhat.com/browse/THREESCALE-9009)
* Skipif doesn't work completely correctly because I cannot filter on minor version

### DIFF
--- a/testsuite/tests/apicast/auth/rhsso/openid_rhsso/test_rhsso_wrong_realm.py
+++ b/testsuite/tests/apicast/auth/rhsso/openid_rhsso/test_rhsso_wrong_realm.py
@@ -1,9 +1,6 @@
 """Tests that accessing APIcast with token from a different RHSSO realm does not work"""
 import pytest
 
-from packaging.version import Version  # noqa # pylint: disable=unused-import
-from testsuite import TESTED_VERSION  # noqa # pylint: disable=unused-import
-
 from testsuite.rhsso import Token
 from testsuite.utils import blame
 
@@ -50,8 +47,6 @@ def backends_mapping(custom_backend):
     return {"/test": custom_backend("backend")}
 
 
-# It doesn't work on 2.13.1
-@pytest.mark.skipif("TESTED_VERSION < Version('2.12')")
 @pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-9009")
 def test_wrong_realm_auth(api_client, wrong_realm_token, correct_realm_token):
     """Using auth from a different RHSSO realm should fail"""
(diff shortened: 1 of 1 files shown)


mganisin (dismissed):

testsuite/tests/apicast/auth/rhsso/openid_rhsso/test_rhsso_wrong_realm.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.fixture(scope="module")
+def backends_mapping(custom_backend):
+    """
+    :return: dict in format {path: backend}
+    """
+    return {"/test": custom_backend("backend")}
+
+
+# It doesn't work on 2.13.1
+@pytest.mark.skipif("TESTED_VERSION < Version('2.12')")
```
mkudlej, line 54: It is not necessary but I think that @pehala can consider to split this test into 2: one for calls which should fail and another one for passing calls.

pehala: I cannot filter by minor, TESTED_VERSION currently contains only `2.12`

mkudlej: Sorry, I've marked wrong line. I've meant that this whole test can be split into passing(20X) asserts and failing(4XX) asserts/http calls.

mganisin: Actually you can and should define patch release @pehala

azgabur (commented):
When I tested without fixture `backends_mapping` and also without `disable_retry_status_list={404}` I encountered `404` in last assert. So I guess they are important. In that case disregard that feedback.

testsuite/tests/apicast/auth/rhsso/openid_rhsso/test_rhsso_wrong_realm.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.fixture(scope="function")
+def correct_realm_token(rhsso_service_info, application):
+    """Token for a correct realm"""
+    return rhsso_service_info.access_token(application)
+
+
+@pytest.fixture(scope="module")
+def backends_mapping(custom_backend):
```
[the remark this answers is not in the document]
pehala: It is actually called somewhere (notably in root conftest), this is just an override. For this specifically it enforced routing policy to be applied at the APIcast level, triggering the bug

testsuite/tests/apicast/auth/rhsso/openid_rhsso/test_rhsso_wrong_realm.py
```diff
@@ -0,0 +1,63 @@
...
+        protocol="openid-connect",
+        standardFlowEnabled=False)
+
+    username = testconfig["rhsso"]["test_user"]["username"]
+    password = testconfig["rhsso"]["test_user"]["password"]
+    realm.create_user(username, password)
+    return client.oidc_client, username, password
+
+
+@pytest.fixture(scope="function")
```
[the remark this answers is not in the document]
pehala: It can, lower scope is always better (if it doesn't significantly affect test length)

mganisin (approved):

dhlavac (approved):

PR #522, opened 2023-03-05
retry instead of sleep in conftest

mganisin:
sleep shouldn't be used to mitigate delays in the app, it isn't reliable
enough. Using backoff is a practive proven to be correct.

### DIFF
--- a/testsuite/resilient.py
+++ b/testsuite/resilient.py
@@ -54,3 +54,9 @@ def accounts_create(client, params):
             client.accounts.delete(client.accounts.read_by_name(params["name"]).entity_id)
             time.sleep(2)
         raise err
+
+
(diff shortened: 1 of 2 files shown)


mdujava (approved):

testsuite/resilient.py
```diff
@@ -54,3 +54,9 @@ def accounts_create(client, params):
             client.accounts.delete(client.accounts.read_by_name(params["name"]).entity_id)
             time.sleep(2)
         raise err
+
+
+@backoff.on_exception(backoff.fibo, ApiClientError, max_tries=8, jitter=None)
```
pehala, line 59: I would like more if the `max_tries` was lower to "Fail faster" in case the 3scale is down, but I am ok with the current implementation too.

pehala (approved):

PR #527, opened 2023-03-07
Add nopersistence mark for generic http tests

JaurbanRH:


pehala (dismissed):

testsuite/tests/system/test_generic_http.py
```diff
@@ -8,7 +8,8 @@
 from testsuite.utils import blame
 
 
-pytestmark = pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-2665")
+pytestmark = [pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-2665"),
+              pytest.mark.nopersistence]
```
mganisin, line 12: explanatory comment is missing

mganisin: @JaurbanRH ?

mganisin (approved):

pehala:
While this probably works, I would add this entire test to nopersistence. This test doesn't make sense to run with persistence as it relies on new application creation. It seems the other tests pass, but it must be from caching something that should not be cached.

pehala (approved):

PR #554, opened 2023-03-20
Adding logs for webdriver session handler

jsmolar:


dhlavac:
What is the purpose of this PR ?  IMHO all this info is present ad logged just we need to change log level

PR #561, opened 2023-03-28
Add check of product version for RHOAM

dhlavac:


testsuite/capabilities/providers.py
```diff
@@ -4,7 +4,7 @@
 from testsuite.configuration import openshift
 
 
-def _rhoam():
+def is_rhoam():
```
mganisin, line 7: this should be moved to dynaconf_loader.py where `data["threescale"]["deployment_type"]` should be set to one of `(onprem, rhoam, saas)`. Such solution has an advantage in posibility to overwrite it by user with `_3SCALE_TESTS_threescale__deployment_type=rhoam`

dhlavac: @mganisin moved to dynaconf loader, for now, there is detection only for RHOAM, do you have any suggestions about how to detect saas environment?

mganisin (approved):

pehala (approved):
I think the deployment_type should probably be an Enum, but this is fine

mganisin:
@dhlavac merge conflict!

pehala (approved):

PR #567, opened 2023-04-03
Remove deprecated rhsso args

mganisin:


pehala:
Since this change is only in the 2.13 version, I change the minimal version we require for python-keycloak in Pipfile

pehala (approved):

JaurbanRH (approved):

PR #569, opened 2023-04-03
Cache them sometimes

pehala:
We already use `cached_property`, so why not use it everywhere? I believe all of these properties are in fact static and should not change, but please correct me if I am wrong


mganisin:
thinking about this all the day, it's probably good idea and safe. Generally speaking it looks dangerous as it creates singletons in fact, what could lead to unexpected behavior.

mganisin (approved):

PR #571, opened 2023-04-04
Wait more on data to be visible in prometheus

mdujava:
in Backend listener internal api test, data is not shown right on next scrape. Waiting 60 seconds makes it more reliable.


pehala:
Is that correct behaviour?

mdujava:
It's not a correct behavior, in first scrape metric will only grow by 4-6 instead of expected 10. This problem might be magnified by usage of the Openshift user workload monitoring. Which contains more components than standalone prometheus compared to previous setup.

mganisin:
@mdujava does it mean that it requires backoff instead of fixed time?

mganisin (approved):

pehala (approved):

PR #573, opened 2023-04-11
Unify operator and subscription methods in OpenShfitClient

pehala:
- All methods are now properties
- All methods now return APIObject or None
  - All the methods that previously returned Selector were always called with `.object()`, and only sometimes try-excepted the potential OpenShiftException when it didn't exist. Now, they will never throw exceptions.
- Change the naming of these methods to reflect their current state better.

This is a breaking change

### DIFF
--- a/testsuite/openshift/client.py
+++ b/testsuite/openshift/client.py
@@ -214,52 +214,54 @@ class OpenShiftClient:
         objects = self.do_action("process", [source, opt_args]).out()
         self.create(objects)
 
-    def get_operator(self):
+    @property
+    def threescale_operator(self):
         """
-        Gets the selector for the 3scale operator
+        Gets the 3scale operator
 
-        :return: the operator pod
+        :return: the operator pod APIObject
         """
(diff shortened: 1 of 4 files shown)


testsuite/openshift/client.py
```diff
@@ -214,54 +214,52 @@ def new_app(self, source, params: Dict[str, str] = None):
...
 
-        :return: the operator pod
+        :return: the operator pod APIObject
         """
         def select_operator(apiobject):
             return apiobject.get_label("com.redhat.component-name") == "3scale-operator" \
                 or apiobject.get_label("rht.subcomp") == "3scale_operator"
 
-        return self.select_resource("pods", narrow_function=select_operator)
+        return self.select_resource("pods", narrow_function=select_operator).object(ignore_not_found=True)
```
mganisin, line 228: This (and others) should raise error

mdujava (approved):

mganisin (approved):

PR #575, opened 2023-04-12
Cleanup skipping implemented on pytest level

mganisin:
Generic cleanup skipping without a need to explicitly handle it in
fixtures. On the other hand uses private access == unstable interface
that can stop working one day.

### DIFF
--- a/testsuite/tests/apicast/parameters/auth/test_rhsso_wrong_realm.py
+++ b/testsuite/tests/apicast/parameters/auth/test_rhsso_wrong_realm.py
@@ -25,8 +25,7 @@ def user_wrong_realm(rhsso_service_info, request, testconfig):
     """User in a wrong realm"""
     realm = rhsso_service_info.rhsso.create_realm(blame(request, "realm2"), accessTokenLifespan=24 * 60 * 60)
 
-    if not testconfig["skip_cleanup"]:
-        request.addfinalizer(realm.delete)
+    request.addfinalizer(realm.delete)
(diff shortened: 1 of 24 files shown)


pehala (commented):
I do not like this approach.
* It adds more magic done in the background
* It switches current behaviour and applies it globally
* It relies on pytest internals and the code is complicated
* There are some finalizers/yields you always want to execute regardless of skip_cleanup and I do not see a way how to do it
   * e.g. api_client/prod_client, where you always want to close the client and clean sessions.

mganisin:
> I do not like this approach.
> 
>     * It adds more magic done in the background

It doesn't, the discovery of this included some magic, the result is free of any incantation.

> 
>     * It switches current behaviour and applies it globally

It does it right.

> 
>     * It relies on pytest internals and the code is complicated

True, it relies on pytest internals and it can break in future, real problem (like with persistence, with which it this is supposed to be used primarily/exclusively; actually can be implemented as part of persistence). However the code is simple and crystal clear (stack structure is not, can be documented), placed on one place and less than 50 lines long.

> 
>     * There are some finalizers/yields you always want to execute regardless of skip_cleanup and I do not see a way how to do it

Maybe... (thinking about this)
>       
>       * e.g. api_client/prod_client, where you always want to close the client and clean sessions.

Personally I did never care this particular one. :-) Nowadays I am more mature and I understand that it can cause slight troubles on server side, however still do not care. :-)

PR #579, opened 2023-04-14
Posibility to define all tools via env

mganisin:
Some tools can be defined with :port or +extra flag. These chars can't
be part of env variable name. Therefore some "magic" has to happen to
allow these options be passed via env. Simple substitutions /+/_plus_/
and /:/_port_/ are made when searching for value.

### DIFF
--- a/config/settings.yaml.tpl
+++ b/config/settings.yaml.tpl
@@ -68,12 +68,22 @@ default:
         webdriver: "" #chrome , firefox or edge(edge with remote drivers)
         remote_url: "" #URL and port to remote selenium instance e.g. http://127.0.0.1:4444
     tools:
-      # tools is a fixture to provide testenv services like echo_api, jaeger and services that are needed for testing
-      # each service is identified by a key, for compatibility reasons some keys are predefined as they have been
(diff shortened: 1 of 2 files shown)


mdujava (approved):

averevki (commented):
Just to be clear: these characters can also be part of the key in the testsuite settings, so simple quotes in environment variables won't be enough, right?

averevki:
Also shouldn't this feature be documented anywhere?

mganisin:
> Just to be clear: these characters can also be part of the key in the testsuite settings, so simple quotes in environment variables won't be enough, right?

I am not sure whether I understand the question.

averevki:
> > Just to be clear: these characters can also be part of the key in the testsuite settings, so simple quotes in environment variables won't be enough, right?
> 
> I am not sure whether I understand the question.

I wasn't sure if those variables with special characters can also be keys and not only values in the settings (in values these special characters usually can be achieved by quotes around the value e.g. `httpbin: "httpbin+svc:8080"` (same in the env variables)), as there are no any examples of this feature in the settings template.

But from the implementation, I see that these special characters can also be used as keys names e.g. 
`httpbin+svc:8080: something`.

averevki (approved):

PR #589, opened 2023-04-19
Add backoff on selenium timeout when opening new admin page

jsmolar:


JaurbanRH (dismissed):

testsuite/ui/webdriver.py
```diff
@@ -30,7 +30,7 @@ def install(self):
         if self.source == 'local':
             self.binary_path = ChromeDriverManager().install()
 
-    def start_session(self):
+    def start_browser(self):
```
dhlavac, line 33: I would prefer to leave there session instead of the browser to avoid misunderstanding. Browser has webdriver and webdriver has browser means we browser.webdriver.browser. which is not a browser but webdriver session.

dhlavac (changes requested):

testsuite/ui/webdriver.py
jsmolar: The original was even more insane. Browser has session and webdriver. Webdriver had another session.

jsmolar: The ideal name would be `selenium` as this is not session as well

dhlavac: session/selenium is ok for me, it is up to you

testsuite/tests/ui/conftest.py
```diff
@@ -77,14 +80,18 @@ def custom_admin_login(navigator, browser):
     Returns parametrized Login fixture for Admin portal.
     To remove cookies and previous login session (from Admin portal), set `fresh` flag to True.
     :param navigator: Navigator Instance
+    :param browser: Browser instance
     :return: Login to Admin portal with custom credentials
     """
 
+    @backoff.on_exception(backoff.constant, InvalidSessionIdException, max_tries=2, jitter=None,
+                          on_backoff=lambda _: browser.restart_session())
```
mganisin, line 88: did you try just `on_backoff=browser.restart_session`

mganisin (approved):

JaurbanRH (approved):

dhlavac (approved):

PR #595, opened 2023-04-20
Fix api docs tests

JaurbanRH:


mganisin (approved):

pehala (approved):

testsuite/ui/widgets/__init__.py
```diff
@@ -346,6 +346,15 @@ class ThreescaleSelect(Select):
     BUTTON_LOCATOR = "./div/button"
 
 
+class APIDocsSelect(Select):
+    """Specific select for 3scale API doc page"""
+
+    def item_select(self, item):
+        element = self.browser.element("//button[@aria-label='Options menu']")
```
dhlavac, line 353: Please add a note that this is for opening the menu because the locator matches all buttons and Select does work

dhlavac (changes requested):

dhlavac (approved):

mganisin (approved):

PR #611, opened 2023-04-28
Reworked Mailhog client for malis and messages tests

dhlavac:


testsuite/mailhog.py
```diff
@@ -80,3 +118,19 @@ def delete(self, mail_id=None):
         else:
             for mail in mail_id:
                 self.request(method="DELETE", endpoint=f"api/v1/messages/{mail}")
+
+    @backoff.on_exception(backoff.fibo, AssertionError, max_tries=8, jitter=None)
+    def assert_message_received(self, expected_count=1, subject=None, content=None):
```
mdujava, line 123: Specify if `subject` and `content` are mutually exclusive or not, as now result of find_by_subject would be overwritten and not checked.

testsuite/mailhog.py
```diff
@@ -60,6 +66,19 @@ def _all_messages(self, start):
             count = response["count"]
         return response
 
+    def delete_searched_messages(self):
```
mganisin, line 69: My understanding is that this is a replacement of a "cleanup" already included in the test. If so i would rather suggest to abandon it. It wasn't best idea to implement in that way. Anyway mailhog requires different cleanup as there is lots of emails generated during execution of all tests. So in my opinion it is absolutely fine to remove that cleanup from the test without replacement.

testsuite/mailhog.py
```diff
@@ -68,8 +87,27 @@ def messages(self, start: int = 0, limit: int = 25):
         response = self.request(params=params, endpoint="api/v2/messages")
         # The quotes are in the returned json escaped by two backslashes and python
         # can not parse it, this replacements fixes it
-        return json.loads(
-            response.content.decode('utf8').replace('\\\\"', '\\"'))
+        return json.loads(response.content.decode('utf8').replace('\\\\"', '\\"'))
+
+    def find_by_subject(self, subject):
```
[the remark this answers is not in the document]
mganisin: Actually let's merge this first, I'll submi my PR after that.
So scratching this comment.
Please reflect mdujava's comment to make this ready to merge

testsuite/mailhog.py
```diff
@@ -26,50 +28,84 @@ def __init__(self, openshift: 'OpenShiftClient',
...
 
-    def request(self, method: str = "GET", params=None, endpoint: str = None):
+    def request(self, method: str = "GET", params: dict = None, endpoint: str = None):
         """Requests the mailhog API"""
         params = params or {}
 
-        full_url = self._url + "/" + endpoint
-        response = requests.request(method=method, url=full_url,
-                                    params=params, verify=False)
+        if not hasattr(self, "_session"):
```
pehala, line 50: Why is this condition needed? Isn´t the session always created in `__init__`?

testsuite/mailhog.py
```diff
@@ -26,50 +28,84 @@ def __init__(self, openshift: 'OpenShiftClient',
...
+
+    def get_messages_by_chunk(self, chunk_size: int = 250, limit: Optional[int] = None):
+        """
+        A generator function that retrieves all MailHog messages in chunks.
+        @param chunk_size: The number of messages to retrieve in each chunk (default 100).
+        @param limit: limits number of returned messages
+        :yield: A list of MailHog messages.
+        """
+        # Retrieve the total number of messages
+        response = self.request(endpoint="/api/v2/messages?count=0")
```
pehala, line 80: you should probably specify count as a param instead

testsuite/mailhog.py
dhlavac: Should be, but from time to time it fails during test run (mostly during long debugging) and this ensures there will be always

PR #618, opened 2023-05-04
Added tests for bulk messages in UI 

dhlavac:
Need to be reviewed with Alpha due to locator changes

### DIFF
--- a/testsuite/mailhog.py
+++ b/testsuite/mailhog.py
@@ -2,7 +2,7 @@
 This module contains wrapper for the Mailhog API
 """
 
-from typing import List, Optional
+from typing import Set, Optional
 import backoff
 import pytest
 import requests
@@ -36,7 +36,7 @@ class MailhogClient:
             else:
                 warn_and_skip("Can't find mailhog, skipping mailhog related tests")
 
-        self._searched_messages_ids: List[str] = []
+        self._searched_messages_ids: Set[str] = set()
         self._session = requests.Session()
 
     @property
@@ -48,9 +48,6 @@ class MailhogClient:
         """Requests the mailhog API"""
         params = params or {}
 
-        if not hasattr(self, "_session"):
-            self._session = requests.Session()
-
         full_url = f"{self._url}/{endpoint}"
         response = self._session.request(method=method, url=full_url, params=params)
         assert response.status_code == 200, f"The request to mailhog failed: {response.status_code} {response.text}"
@@ -60,7 +57,7 @@ class MailhogClient:
         """Append message ID to list for cleanup
         @param message: message to be cleaned
         """
-        self._searched_messages_ids.append(message["ID"])
+        self._searched_messages_ids.add(message["ID"])
 
     def all_messages(self):
         """Return all messages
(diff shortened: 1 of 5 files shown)


testsuite/tests/ui/mail_and_messages/test_messages.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    subject = blame(request, 'subject')
+    content = blame(request, 'message')
+    settings = navigator.navigate(ApplicationsView)
+    app_count = settings.send_email_to_all_apps(subject, content)
+
+    custom_devel_login(account)
+    messages_page = navigator.navigate(InboxView)
+
+    assert len(list(filter(lambda msg: msg['SUBJECT'] == subject, messages_page.messages_table.read()))) != 0
```
pehala, line 25: I would instead use list comprehension (just a generator will suffice)

testsuite/ui/views/admin/audience/application.py
```diff
@@ -1,21 +1,59 @@
...
+    send_email_btn = GenericLocatorWidget(".//button[text()='Send email']")
+    email_window = View.nested(ApplicationsBulkEmailWindow)
+
+    def send_email_to_all_apps(self, subject=None, body=None):
+        """Send message to all selected applications
+        @param subject: message subject
+        @param body: message body
+        @return number of selected application (to verify correct number of messages) """
+        self.all_app_checkbox.check()
+        self.send_email_btn.wait_displayed(delay=3)
```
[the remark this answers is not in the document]
dhlavac: Yes, it should succeed on a very slow 3scale, backoff there doesn't make sense because it only waits for elements to proceed in next interactions.

testsuite/mailhog.py
```diff
@@ -47,7 +47,7 @@ def request(self, method: str = "GET", params: dict = None, endpoint: str = None
         """Requests the mailhog API"""
         params = params or {}
 
-        if not hasattr(self, "_session"):
+        if not self._session:
```
pehala, line 50: I think this is an unnecessary change, I do not think this improves anything, and I still do not see a reason for this condition. However, I will leave this up to @mganisin if this change should be here or not

mganisin: More importantly these statements are not equal, this is change in behavior. If I remember this might be switch back to original statement. Comment (here in PR) may help to understand tihs.

dhlavac: It checks if a session still exists during request method call. From time to time (during debugging happens that there is no session - None returned) should I remove it?

testsuite/mailhog.py
```diff
@@ -60,7 +60,8 @@ def append_to_searched_messages(self, message):
         """Append message ID to list for cleanup
         @param message: message to be cleaned
         """
-        self._searched_messages_ids.append(message["ID"])
+        if message["ID"] not in self._searched_messages_ids:
```
mganisin, line 63: use `set()` if you want unique items

testsuite/tests/ui/mail_and_messages/test_messages.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    subject = blame(request, "subject")
+    content = blame(request, "message")
+    settings = navigator.navigate(ApplicationsView)
+    app_count = settings.send_email_to_all_apps(subject, content)
+
+    custom_devel_login(account)
+    messages_page = navigator.navigate(InboxView)
+
+    assert len([msg for msg in messages_page.messages_table.read() if msg["SUBJECT"] == subject]) != 0
```
mganisin, line 25: if `blame` is in use, maybe len/count is known exactly

pehala (approved):

JaurbanRH (approved):

PR #624, opened 2023-05-09
If rerun is defined, only rerun failed tests

mdujava:


pehala (approved):

Makefile
```diff
@@ -31,6 +31,10 @@ ifdef html
 PYTEST += --html=$(resultsdir)/report-$(@F).html --self-contained-html
 endif
 
+ifdef rerun
+PYTEST += --last-failed --last-failed-no-failures none
```
mganisin, line 35: No. :-)

`--last-failed` is pointless in my opinion `make smoke rerun=1` or `make smoke flags=--lf` are equivalent, latter is more explicit in what happens and doesn't introduce new semantics (wins twice).

`--last-failed-no-failures none` is pretty useful, great to be in pytest.ini

What do you think?

mganisin (approved):

pehala (approved):

PR #626, opened 2023-05-10
Fix test_devel_sections

JaurbanRH:


mganisin (approved):

testsuite/tests/ui/devel/test_devel_sections.py
```diff
@@ -81,6 +78,7 @@ def cleanup():
 @pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-9020")
 @pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-836")
 @pytest.mark.skipif("TESTED_VERSION < Version('2.14')")
+@pytest.mark.usefixtures("login", "dev_portal_group")
```
jsmolar, line 81: Why are we using usefixture "login", when the first action we take in the test is `custom_devel_login(account=account)`

jsmolar (approved):

mganisin (approved):

PR #629, opened 2023-05-10
TLS tests belong to sandbag

mganisin:


pehala (approved):
Do we run sandbag UI tests?

mganisin:
> Do we run sandbag UI tests?

Yes, on rhoam, there is no special target because it is pretty special (and maybe even temporrary) but sometimes (exclusively from automation) this is handy (already soma sandbags in UI).
No manual execution requirement is main reason for no need of special target (`ui flags=--drop-sandbag` works fine for automation as well as for very rare manual execution)

JaurbanRH (approved):

PR #632, opened 2023-05-11
Updated api docs page object to OAS3

jsmolar:
3scale API Docs are now using OAS3. In addition, all OAS3 page objects now use the same widgets for endpoint representation.
Why we need this PR:

1. OAS3 was used for 3scale API docs (old 3scale design was completely removed)
2. Our OAS3 object used plane selenium (please see deleted code in `class ActiveDocV3Section(Widget)`)
3. Our new `Endpoint` is ParametrizedView, which is an ideal tool from Widgetastic to identify multiple same objects (views) on the same page. By doing this the usage is simple and straightforward:  `view.oas3.endpoint("GET", "/pets").execute({"user_key": key})`
4. Point 3 brought many benefits of Widgetastic to our OAS3 pages. The one that stands out, is that we no longer need so many backoffs for this View.
5. New`Ednpoint`View is the general solution. It brings a one-size-fits-all solution to all the current and future OAS3 UI tests

### DIFF
--- a/testsuite/tests/ui/api_docs/test_autocomplete.py
+++ b/testsuite/tests/ui/api_docs/test_autocomplete.py
@@ -1,7 +1,6 @@
 """Rewrite of spec/ui_specs/autocomplete_spec.rb"""
 import pytest
 
-from testsuite import rawobj
 from testsuite.ui.views.admin.settings.api_docs import APIDocsView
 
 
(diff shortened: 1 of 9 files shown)


testsuite/tests/ui/api_docs/test_autocomplete.py
```diff
@@ -18,14 +17,14 @@ def test_autocomplete(navigator, service, testconfig):
...
+    endpoint = api_docs_page.endpoint("GET", "/admin/api/services/{id}.xml")
+    endpoint.set_param("access_token", token)
+    endpoint.set_param("id", service["name"])
 
-    product_id_element = api_docs_page.get_id_input_by_name(service["name"])
-    auto_comp_prod_id = product_id_element.text
-    product_id_element.click()
-    assert auto_comp_prod_id == str(service["id"])
+    # endpoint.get_param("id").click()
+    # product_id_element = api_docs_page.get_id_input_by_name(service["name"])
```
JaurbanRH, line 25: I think that these shouldn't be here.

testsuite/ui/widgets/oas3.py
```diff
@@ -0,0 +1,117 @@
...
+        self.browser.selenium.find_element(By.XPATH, item).click()
+
+    def __locator__(self):
+        return self.locator
+
+    def __repr__(self):
+        return "{}{}".format(type(self).__name__, self.locator)
+
+
+class Endpoint(ParametrizedView):
```
JaurbanRH, line 32: I think that this view should be located with other views and not with widgets.

jsmolar: The View is Widget. `/ui/wdigets` contains elements used across different page objects across the UI test suite. `class Endpoint` is also designed for multiple page objects. 
It does not make sense to move it to the View package.

testsuite/tests/ui/api_docs/test_autocomplete.py
jsmolar: You are right. We need to discuss if this is the correct approach how to test autocomplete

testsuite/ui/widgets/oas3.py
dhlavac: I agree with Smolar, it makes more sense to have it as a generic endpoint widget as it is present on multiple pages

dhlavac:
Rebase please(missing fix for login in PR)

mganisin (approved):

PR #643, opened 2023-05-24
first WIP PR to enable smoke test with CRD 3scale client

mkudlej:
It introduces APIAP for all smoke tests with related changes.


testsuite/tests/apicast/policy/routing/test_routing_policy_multiple_conditions_or.py
```diff
@@ -36,10 +37,10 @@ def test_routing_policy_route_testing(api_client, private_base_url):
     Test for the request send with Test1 and Test2 to echo/route/get
     """
     parsed_url = urlparse(private_base_url("echo_api"))
-    response = api_client().get("/get", headers={"Test1": "route", "Test2": "testing"})
+    response = api_client().get("/", headers={"Test1": "route", "Test2": "testing"})
```
pehala, line 40: Why?

testsuite/tests/conftest.py
```diff
@@ -374,10 +375,11 @@ def threescale(testconfig, request):
         tenant = custom_tenant()
         return tenant.admin_api(ssl_verify=testconfig["ssl_verify"], wait=0)
 
-    return client.ThreeScaleClient(
+    return client.ThreeScaleClientCRD(
         testconfig["threescale"]["admin"]["url"],
         testconfig["threescale"]["admin"]["token"],
         ssl_verify=testconfig["ssl_verify"],
+        ocp_provider_ref=testconfig["threescale"]["admin"]["account_ref"],
```
mganisin, line 382: I don't know what is `ocp_provider_ref` or `threescale.admin.account_ref` however I am pretty sure `ocp_provider_ref != threescale.admin.account_ref` (scope ot these names ares disjunct). What if they had same name?

Yet "thanks to" `_ref` suffix these names look pretty alien, do not fit to the rest.

testsuite/tests/conftest.py
```diff
@@ -841,10 +850,10 @@ def finalizer():
             if backends:
                 for path, backend in backends.items():
                     svc.backend_usages.create({"path": path, "backend_api_id": backend["id"]})
-                proxy_params = {}
+                proxy_params.pop("api_backend")
+                proxy_params.pop("api_test_path")
```
mganisin, line 854: this is breaking change, will make many tests just crash

testsuite/tests/conftest.py
```diff
@@ -757,12 +759,17 @@ def finalizer():
 
 
 @pytest.fixture(scope="module")
-def backends_mapping():
+def backend_default(private_base_url, custom_backend):
+    return custom_backend("backend_default", endpoint=private_base_url())
```
mganisin, line 763: as far as I undertand this, this is irrelevant change to this PR. Even if necessary, better to have it in separate commit at least, better in separate PR (still difficult to accept, there are explicit tests for APIAP, however there are no explicit tests for legacy behavior, so this is something to discuss as it will reduce test coverage from certain point of view)

testsuite/rawobj.py
```diff
@@ -85,7 +90,12 @@ def Application(
     if description is None:
         description = f"application {name}"
 
-    obj = {"name": name, "plan_id": application_plan["id"], "description": description}
+    obj = {
+        "name": name,
+        "plan_id": application_plan["id"],
+        "description": description,
+        "service_id": application_plan.service["id"],
```
mganisin, line 97: these modifications of rawobj are all suspicious, we will see

Pipfile
```diff
@@ -61,5 +61,8 @@ pytest-metadata = ">=3.0.0"
 pytest-html = "*"
 ansi2html = "*"
 
+# 3scale Capabilities API client 
+3scale-api-crd = {git = "https://github.com/mkudlej/3scale-api-python-crd.git", editable = true}
```
JaurbanRH, line 65: Make release of 3scale-api-crd instead of using git url.

testsuite/mockserver.py
```diff
@@ -83,5 +83,5 @@ def verify_sequence(self, expected_requests) -> bool:
             verify=self.verify,
         )
         if response.status_code == 400:
-            raise HTTPError("Invalid matcher format", response=response)
+            raise HTTPError("Invalid matcher format", response=response)  # type: ignore
```
JaurbanRH, line 86: Is this necessary?

dhlavac:
@mkudlej please add mention about capabilities tests to the documentation for the next generations :)

testsuite/tests/conftest.py
```diff
@@ -51,6 +53,13 @@ def term_handler():
     signal.signal(signal.SIGTERM, orig)
 
 
+@pytest.fixture(scope="session")
+def api_client_type(pytestconfig):
+    """Do tests use REST API client(False) or CR Capabilities client(True)?"""
```
JaurbanRH, line 58: I would rename this fixture to something like `use_capabilities_client` to match bool return value and for more clear usage later in the code.

PR #652, opened 2023-05-30
PF4 UI changes

jsmolar:


testsuite/ui/views/admin/backend/__init__.py
```diff
@@ -52,7 +52,7 @@ class BaseBackendView(BaseAdminView):
     """
 
     NAV_ITEMS = ["Overview", "Analytics", "Methods and Metrics", "Mapping Rules"]
-    nav = NavigationMenu(id="mainmenu")
+    nav = NavigationMenu(component_id="OUIA-Generated-Nav-1")
```
mganisin, line 55: Comment for next time, not now. There should be something like:

```
class ThrescaleNavigationMenu:  # name should be completely different, not this loooong tragedy, I only wanted to differentiate it from generic NavigationMenu (if appropriate, the name can be NavigationMenu also here)
    def __init__(self, component_id="OUIA-Generated-Nav-1"):
      ...
```
This will prevent duplicate lines and will make this change smaller and simpler. Obviously it is not bound to NavigationMenu, it is applicable generally.

testsuite/ui/widgets/__init__.py
```diff
@@ -130,78 +129,6 @@ def clear_all(self):
...
-    """
-
-    DEFAULT_LOCATOR = './/div[contains(@class, "pf-c-context-selector")]'
-    BUTTON_LOCATOR = './/a[@title="Context Selector"]'
-
-
-# pylint: disable=abstract-method
-# Widget contains fill method which raise not implemented exception if widget is not fillable but pylint detect it as
-# an abstract method
-class NavigationMenu(Navigation):
```
mganisin, line 146: `from ouia import NavigationMenu` here would safe few changes from other files

mganisin (approved):

JaurbanRH (approved):

PR #660, opened 2023-06-05
Make apicast tests comaptible with peristence plugin

JaurbanRH:


testsuite/tests/apicast/test_limit_exceeded_metric.py
```diff
@@ -30,17 +30,24 @@ def application(application):
     return application
 
 
+@pytest.fixture(scope="module")
+def exceed_limit(api_client):
+    """Exceed limit of /anything/exceeded endpoint"""
```
mganisin, line 35: This looks pretty weird, thus requires good description why it should be done this way. As it is part of this PR it looks obvious that this is to have this test working after persistence load, however once in the codebase, it won't be evident and there is no relation to persistence (yet I may be totally wrong and the reason is completely different)

testsuite/tests/apicast/policy/batcher/test_batcher_policy_mapping_rules.py
```diff
@@ -44,7 +44,9 @@ def service(service):
 @pytest.fixture(scope="module")
 def client(api_client):
     """We are testing path that doesn't match mapping rule so we need to disable retry"""
-    assert api_client().get("/get").status_code == 200
+    assert api_client().get("/get").status_code == 200  # Ensures that service is set up correctly
+
+    sleep(50)  # need to wait 50 seconds because of batcher policy to ensure stable environment for tests
```
pehala, line 49: Was this problem before as well, or is the sleep just for persistence?

testsuite/tests/apicast/policy/batcher/test_batcher_policy_mapping_rules.py
```diff
@@ -67,4 +69,4 @@ def test_batcher_policy_append(client, application):
     sleep(50)
 
     usage_after = analytics.list_by_service(application["service_id"], metric_name="hits")["total"]
-    assert usage_after == usage_before + 4
+    assert usage_after == usage_before + 3
```
pehala, line 72: Why was the usage_after changed?

JaurbanRH: Because before the change there were 3 hits from test and one from test setup which is now correctly handled, hence only 3 hits from test left.

testsuite/tests/apicast/policy/batcher/test_batcher_policy_mapping_rules.py
JaurbanRH: It was a problem before as well, but was resolved by +4 in test assert instead of +3 as it should. The persistence plugin just shows us the problem with this test.

pehala (approved):

mganisin (approved):

PR #674, opened 2023-06-13
Remove drop-sandbag from persistence make target

JaurbanRH:
#660 needs to be merged first

### DIFF
--- a/testsuite/openshift/crd/operator.py
+++ b/testsuite/openshift/crd/operator.py
@@ -1,0 +1,26 @@
+"""Module containing Operator object"""
+import openshift as oc
+from openshift import APIObject
+
+
+class Operator(APIObject):
+    """Operator CRD object with custom serializer and deserializer for pickle module"""
+
+    def __getstate__(self):
+        """
+        Custom serializer for pickle module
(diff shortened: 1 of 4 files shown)


testsuite/openshift/crd/operator.py
```diff
@@ -0,0 +1,25 @@
...
+            "name": self.model.metadata.name,
+            "namespace": self.model.metadata.namespace,
+        }
+
+    def __setstate__(self, state):
+        """
+        Custom deserializer for pickle module
+        more info here: https://docs.python.org/3/library/pickle.html#object.__setstate__
+        """
+        result = oc.invoke("get", ["pod", state["name"], "-n", state["namespace"], "-o", "yaml"], no_namespace=True)
```
pehala, line 24: A few notes:
* While this is not necessarily wrong, there is also an option to use selectors:
```
with some_context:
  oc.selector(f"pod/{name}).object_json()
```
* This could be made more generic by persisting `Context` class (`self.context`), which always contains all the information about where to execute the command
* If you won't use context, at least also serialize a type of the openshift object, to make this more generic

PR #677, opened 2023-06-19
Deletion automated mail tests for service and account

dhlavac:


testsuite/tests/system/messages/test_automatic_emails.py
```diff
@@ -0,0 +1,70 @@
...
+
+@pytest.fixture()
+def service_settings(request):
+    """
+    Makes service with required name
+    """
+    return {"name": blame(request, "svc")}
+
+
+@pytest.fixture()
```
mganisin, line 42: why changing the scope?

testsuite/tests/system/messages/test_automatic_emails.py
```diff
@@ -0,0 +1,70 @@
...
+    Create new account
+    Delete account
+    Assert that a delete email has been sent
+    """
+    account = custom_account(params, autoclean=False)
+    account.delete()
+    mailhog_client.assert_message_received(subject=f"Account {account.entity_name} deleted", expected_count=1)
+
+
+def test_service_deleted_automatic_email(mailhog_client, only_service):
```
mganisin, line 61: not necessary to change anything, just a comment. These two tests seem to share nothing. In such case having them in separate files would make much clearer which fixture belongs to which test.

testsuite/tests/system/messages/test_automatic_emails.py
```diff
@@ -0,0 +1,70 @@
...
+import time
+
+import pytest
+
+from testsuite import rawobj
+from testsuite.utils import blame
+
+
+@pytest.fixture()
+def params(request):
```
mganisin, line 14: a fixture for an account is more appropriate than params. Unless account creation is part of test procedure, looking at the test it doesn't seem to be the case

testsuite/tests/system/messages/test_automatic_emails.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    :return: params for custom account
+    """
+    name = blame(request, "id")
+
+    return {"name": name, "username": name, "org_name": name, "email": f"{name}@anything.invalid"}
+
+
+@pytest.fixture()
+def custom_app(custom_application, custom_app_plan, service, lifecycle_hooks, request):
```
mganisin, line 24: This is unused

JaurbanRH (approved):

JaurbanRH (approved):

Issue #159, opened 2022-02-22
Test emails: User activation signup through invitation


mganisin:

dhlavac:
Automated

PR #678, opened 2023-06-19
User invitation mail and signup test

dhlavac:
T est for https://github.com/3scale-qe/3scale-tests/issues/160 and  https://github.com/3scale-qe/3scale-tests/issues/159


testsuite/tests/ui/mail_and_messages/test_automatic_mails.py
```diff
@@ -0,0 +1,47 @@
...
+    password = blame(request, "passwd")
+
+    invitation = navigator.navigate(AccountInvitationNewView, account=account)
+    invitation.invite_user(email)
+
+    # assert inside mailhog method
+    mailhog_client.assert_message_received(
+        subject=f"Invitation to join {account.entity_name}", receiver=email, expected_count=1
+    )
+    mail = mailhog_client.find_message(subject=f"Invitation to join {account.entity_name}", receiver=email)
```
mganisin, line 33: Actually this could be returned by `assert_message_received`. It can be bit unusual for assert to return something, anyway this means same call twice -> twice as much time to run the test.

testsuite/ui/views/devel/login.py
```diff
@@ -113,6 +113,46 @@ def prerequisite(self):
...
+
+    path_pattern = "/signup"
+    username = TextInput(id="user_username")
+    email = TextInput(id="user_email")
+    password = TextInput(id="user_password")
+    password2 = TextInput(id="user_password_confirmation")
+    recaptcha = View.nested(ReCaptcha)
+    skip_wait_displayed = True
+
+    # pylint: disable=too-many-arguments
```
mganisin, line 134: Really? is 4 arguments already too many?

testsuite/tests/ui/mail_and_messages/test_automatic_mails.py
pehala: Or alternatively, you could `find_message` first and assert could work with already fetched mails like this:
```
mail = find_message(...)
assert_message_received(mail, expected_count=1)
```

dhlavac: @mganisin It doesn't mean twice much time as fetching mails is only a little part of the test time, but I am not sure about the correct solution as well

mganisin (approved):

testsuite/ui/views/devel/login.py
```diff
@@ -113,6 +113,45 @@ def prerequisite(self):
...
+        if username:
+            self.username.fill(username)
+        self.password.fill(passwd)
+        self.password2.fill(passwd)
+        if submit:
+            self.signup_button.click()
+
+    def check_recaptcha(self):
+        """
+        Checks if reCaptcha exists and if it does, it execute the recaptcha verification
```
jsmolar, line 147: This does not check if the recaptcha exists.

JaurbanRH (approved):

testsuite/ui/views/devel/__init__.py
```diff
@@ -137,7 +137,6 @@ def is_displayed(self):
         return (
             BaseDevelView.is_displayed.fget(self)
             and self.path in self.browser.url
-            and self.organization.is_displayed
```
jsmolar, line 140: Why do you remove organization? Remove BaseDevelView instead ;)

dhlavac: because signup view is different by signup flow for example via mail invitation there is no organization field and due to inheritance I need to have common base

testsuite/ui/views/devel/login.py
dhlavac: it depends on how you understand "check", if there is no recaptcha there will be an exception about no element.

dhlavac: it depends on how you understand "check", if there is no recaptcha there will be an exception about no element.

jsmolar (approved):

PR #680, opened 2023-06-20
Test for application plan change mail notifications

dhlavac:
Covers https://github.com/3scale-qe/3scale-tests/issues/161

### DIFF
--- a/testsuite/tests/ui/mail_and_messages/test_email_application.py
+++ b/testsuite/tests/ui/mail_and_messages/test_email_application.py
@@ -1,10 +1,10 @@
 """Test of automatic mails functionality connected to applications in UI"""
 import pytest
+from threescale_api.resources import Service
 
 from testsuite import rawobj
 from testsuite.utils import blame
 
-from threescale_api.resources import Service
 from testsuite.ui.views.admin.audience.application import ApplicationDetailView
 from testsuite.ui.views.devel.applications import DevelApplicationDetailView
 
@@ -63,8 +63,6 @@ def test_app_key_delete_mails_notification(app_id_key_service, application, navi
     )
 
 
-
-@pytest.mark.usefixtures("login")
 def test_mail_application_plan_changed(navigator, service, application, extra_plan, mailhog_client):
     """
     Test:
--- a/testsuite/tests/ui/mail_and_messages/test_email_application.py
+++ b/testsuite/tests/ui/mail_and_messages/test_email_application.py
@@ -1,8 +1,12 @@
 """Test of automatic mails functionality connected to applications in UI"""
 import pytest
 
+from testsuite import rawobj
+from testsuite.utils import blame
+
 from threescale_api.resources import Service
 from testsuite.ui.views.admin.audience.application import ApplicationDetailView
+from testsuite.ui.views.devel.applications import DevelApplicationDetailView
 
 pytestmark = pytest.mark.usefixtures("login")
 
@@ -12,6 +16,12 @@ def service_settings(service_settings):
     """Set auth mode to app_id/app_key"""
     service_settings.update({"backend_version": Service.AUTH_APP_ID_KEY})
     return service_settings
+
+
+@pytest.fixture(scope="function")
+def extra_plan(request, service, custom_app_plan):
+    """Additional application plan to choose"""
+    return custom_app_plan(rawobj.ApplicationPlan(blame(request, "extra_plan")), service)
 
 
 @pytest.fixture(scope="module")
@@ -49,5 +59,49 @@ def test_app_key_delete_mails_notification(app_id_key_service, application, navi
     mailhog_client.assert_message_received(
         subject="Application key has been deleted",
         content=f"A key has been deleted from your application {application['name']}.",
+        expected_count=1,
+    )
+
+
+
+@pytest.mark.usefixtures("login")
+def test_mail_application_plan_changed(navigator, service, application, extra_plan, mailhog_client):
+    """
+    Test:
+        - Change application plan
+        - Assert that API consumer is notified about plan change
+        - Assert that API provider is notified about plan change
+    """
+    app = navigator.navigate(ApplicationDetailView, application=application, product=service)
+    app.change_plan(extra_plan["name"])
+
+    mailhog_client.assert_message_received(
+        subject=f"Application plan changed to '{extra_plan['name']}'", expected_count=1
+    )
+    mailhog_client.assert_message_received(
+        subject=f"Application {application.entity_name} has changed to plan {extra_plan['name']}", expected_count=1
+    )
+
+
+# pylint: disable=too-many-arguments
+def test_mail_request_app_plan_changed(account, navigator, custom_devel_login, application, extra_plan, mailhog_client):
+    """
+    Test:
+        - Request to change application plan
+        - Assert that API consumer is notified about request plan change
+        - Assert that API provider is notified about request plan change
+    """
+    custom_devel_login(account=account)
+    app_view = navigator.navigate(DevelApplicationDetailView, application=application)
+    app_view.change_plan_link.click()
+    app_view.change_plan(extra_plan["id"])
+    mailhog_client.assert_message_received(
+        subject=f"Action required: {account.entity_name} from {account.entity_name} requested an app plan change",
+        expected_count=1,
+    )
+    mailhog_client.assert_message_received(
+        subject="Plan change request has been received",
+        content=f"Dear {account.entity_name},\r\n\r\nProvider Name has received your request to have your plan "
+        f"changed to {extra_plan['name']} for application {application.entity_name}.",
         expected_count=1,
     )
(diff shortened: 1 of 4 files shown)


testsuite/tests/ui/mail_and_messages/test_automatic_messages.py
```diff
@@ -0,0 +1,43 @@
+"""Test of automatic mails functionality in UI"""
+import pytest
+
+from testsuite import rawobj
+from testsuite.ui.views.admin.audience.application import ApplicationDetailView
+from testsuite.utils import blame
+
+
+@pytest.fixture(scope="module")
+def custom_app(service, custom_app_plan, custom_application, lifecycle_hooks, request):
```
pehala, line 10: why do you need custom app? There should be one created already

testsuite/tests/ui/mail_and_messages/test_automatic_messages.py
```diff
@@ -0,0 +1,43 @@
...
+@pytest.mark.usefixtures("login")
+def test_mail_application_plan_changed(custom_app_plan, request, navigator, service, custom_app, mailhog_client):
+    """
+    Test:
+        - Change application plan
+        - Assert that API consumer is notified about plan change
+        - Assert that API provider is notified about plan change
+    """
+    application = custom_app()
+    app_plan = custom_app_plan(rawobj.ApplicationPlan(blame(request, "app_plan")), service)
```
mganisin, line 34: this could be a fixture

testsuite/tests/ui/mail_and_messages/test_automatic_mails.py
```diff
@@ -44,3 +67,48 @@ def test_invitation_mail(account, navigator, mailhog_client, provider_account, r
     custom_devel_login(name=username, password=password)
 
     assert "SIGNED IN SUCCESSFULLY" in browser.element('//*[@id="flash-messages"]/div/div').text
+
+
+# pylint: disable=too-many-arguments
```
mganisin, line 72: and unnecessary here

testsuite/tests/ui/mail_and_messages/test_automatic_mails.py
```diff
@@ -44,3 +67,48 @@ def test_invitation_mail(account, navigator, mailhog_client, provider_account, r
...
+
+    mailhog_client.assert_message_received(
+        subject=f"Application plan changed to '{extra_plan['name']}'", expected_count=1
+    )
+    mailhog_client.assert_message_received(
+        subject=f"Application {application.entity_name} has changed to plan {extra_plan['name']}", expected_count=1
+    )
+
+
+def test_mail_request_app_plan_changed(extra_plan, account, navigator, custom_devel_login, custom_app, mailhog_client):
```
mganisin, line 93: but `# pylint: disable=too-many-arguments` required here, not visible because of a "problem" with line 34

testsuite/tests/ui/mail_and_messages/test_automatic_mails.py
```diff
@@ -2,12 +2,35 @@
...
+@pytest.fixture(scope="module")
+def extra_plan(request, service, custom_app_plan):
+    """Additional application plan to choose"""
+    return custom_app_plan(rawobj.ApplicationPlan(blame(request, "extra_plan")), service)
+
+
+@pytest.fixture(scope="module")
+def custom_app(service, custom_app_plan, custom_application, lifecycle_hooks, request):
+    """
+    Create an application with predefined application plans
```
mganisin, line 22: This is insufficient comment (many other are, different people, different PRs, but here it is more important). This doesn't explain why it is needed and how it is used. Better would be that application is created during test itself and therefore this is needed.

Anyway I'd rather suggest different approach. To plans, one for each test, and one app (main application fixture) sound better to me. On the other hand this proposed solution would work if upgrade testing using pytest_persistence would be in use, but a model with two plans would not. However that is more a problem of pytest_persistence and less of the design.

That being written I am not requesting the change to two plans, that's something to think about. I am requesting the change of docstring with no doubt.

dhlavac: Finally, I remember why I do it this way and I absolutely agree with you that your proposed flow makes much more sense.  But, the implementation of a developer portal is not optimal, and when a single application has more than two plans active we would need to extend some Widgets locators and usage of some of our base elements because each plan has its own buttons that are not generated dynamically, but only overlayed one by another. I will try to change it to your proposed solution.

testsuite/ui/widgets/buttons.py
```diff
@@ -23,25 +23,33 @@ class ThreescaleButton(widgetastic_patternfly4.Button):
     therefore a class with custom locator is needed.
     """
 
-    def __init__(self, parent, text, classes, **kwargs):
+    def __init__(self, parent, classes, text=None, element_id=None, **kwargs):
```
mganisin, line 26: Order of commits is reversed, the commit with update of button should be first. Actually even separate PR would be nice however it is not necessary. Certainly all the button relevant changes should be in one commit (not the case now).

According to logic of widgetastic_patternfly4.Button `element_id` should be just `id` actually (pylint may complain, I don't know, anyway if so, it's good to mute it). Actually, again according to behavior of Button superclass, `id` does not have to be listed in args, all the kwargs should be part of locator or gather `elif 'id' in kwargs:` instead of `elif element_id`.

This signature will not break the interface (less lines to change):
```
def __init__(self, parent, text=None, *, classes, **kwargs):
```

and still should be compatible with both new and old usage (unless `classes` arg is used somewhere without keyword; I didn't find anything like this, anyway even in such case changing that usage would be more preferable than current proposal as it would preserve interface of supeclass Button)

testsuite/tests/ui/mail_and_messages/test_automatic_mails.py
```diff
@@ -2,14 +2,13 @@
 import re
 import pytest
 
-
```
mganisin, line 5: I know that this does not matter, anyway this is unnecessary change. (is it manual or pycharm?)

dhlavac: black did it due to moving the test from this file

testsuite/ui/widgets/buttons.py
```diff
@@ -25,9 +25,14 @@ class ThreescaleButton(widgetastic_patternfly4.Button):
 
     def __init__(self, parent, text, classes, **kwargs):
         base = "(self::a or self::button or (self::input and (@type='button' or @type='submit')))"
-        classes = " and ".join(f"contains(@class, {quote(i)})" for i in classes)
-        conditions = f"(({classes}) or contains(text(), {quote(text)}))"
-        super().__init__(parent, locator=f".//*[{base} and {conditions}]", **kwargs)
+        class_conditions = " and ".join(f"contains(@class, {quote(cls)})" for cls in classes)
+
+        conditions = f"({class_conditions} or contains(text(), {quote(text)}))"
+        if "id" in kwargs:
```
pehala, line 31: why not specify id as a optional parameter?

mganisin: @pehala this particular line is result of previous review, should be clear from that

testsuite/tests/ui/mail_and_messages/test_email_application.py
```diff
@@ -51,3 +61,45 @@ def test_app_key_delete_mails_notification(app_id_key_service, application, navi
...
+    )
+    mailhog_client.assert_message_received(
+        subject=f"Application {application.entity_name} has changed to plan {extra_plan['name']}", expected_count=1
+    )
+
+
+# pylint: disable=too-many-arguments
+def test_mail_request_app_plan_changed(account, navigator, custom_devel_login, application, extra_plan, mailhog_client):
+    """
+    Test:
```
jsmolar, line 87: This doc string is the same for both tests in this module.

testsuite/tests/ui/mail_and_messages/test_email_application.py
```diff
@@ -51,3 +61,45 @@ def test_app_key_delete_mails_notification(app_id_key_service, application, navi
...
+        - Change application plan
+        - Assert that API consumer is notified about plan change
+        - Assert that API provider is notified about plan change
+    """
+    custom_devel_login(account=account)
+    app_view = navigator.navigate(DevelApplicationDetailView, application=application)
+    app_view.change_plan_link.click()
+    app_view.change_plan(extra_plan["id"])
```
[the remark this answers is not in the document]
dhlavac: It cannot be done because one is request to change and second is change via admin portal

testsuite/ui/views/devel/applications.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def prerequisite(self):
+        return Navbar
+
+    @property
+    def is_displayed(self):
+        return self.path in self.browser.url
+
+
+class DevelApplicationDetailView(BaseDevelView):
```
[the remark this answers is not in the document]
dhlavac: In that case, we will need to rename it in tests because there are used in the same file as  ApplicationDetailView from admin portal

jsmolar (approved):

mganisin (approved):

PR #696, opened 2023-06-30
Mark zync annotation tests as nopersistence

mdujava:


JaurbanRH (approved):

testsuite/tests/prometheus/zync/test_annotations.py
```diff
@@ -10,6 +10,7 @@
 
 pytestmark = [
     pytest.mark.sandbag,  # requires openshift
+    pytest.mark.nopersistence,  # fixture saves pod name, which changes with pod redeployment
```
dhlavac, line 13: Shouldn't it just save something that matches the pod basename - IMHO there is always some base

pehala: If you are testing different pod, it is not a persistence test, you are checking the new annotations on a new pod, which is covered by a normal run instead.

mganisin (approved):

PR #699, opened 2023-07-04
Fix: Skip cleanup of config map if requested

mdujava:
this test was failing on persistence-load due to missing ConfigMap


pehala (approved):

mdujava:
Not needed after https://github.com/JaurbanRH/pytest-persistence/pull/5/ is merged.

PR #704, opened 2023-07-10
New widgets to handle (new) search elements

mganisin:
Search is shared and (more importantly) has changed. Here is update
according to new layout plus widgets to cover it.

### DIFF
--- a/testsuite/ui/views/admin/backend/__init__.py
+++ b/testsuite/ui/views/admin/backend/__init__.py
@@ -1,10 +1,11 @@
 """Essential Views for Backends Views"""
 from widgetastic_patternfly4 import PatternflyTable
-from widgetastic.widget import Text, TextInput, GenericLocatorWidget
+from widgetastic.widget import Text
 
 from testsuite.ui.navigation import step
 from testsuite.ui.views.admin.foundation import BaseAdminView
 from testsuite.ui.widgets.ouia import Navigation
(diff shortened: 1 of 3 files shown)


testsuite/ui/widgets/search.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class SearchInput(TextInput):
+    """Widget representing 3scale search input used on several places"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, locator="//input[@aria-label='Search input']", **kwargs)
+
+
+class SearchButton(GenericLocatorWidget):
```
dhlavac, line 16: Please move the search button to buttons module and ideally Input to general widget module

dhlavac (changes requested):

testsuite/ui/widgets/search.py
mganisin: Or alternatively, search related items could be bundled together. None of these is supposed to live without the other one.

jsmolar:
I was thinking about this for a long time. These two buttons, should be one widget: https://www.patternfly.org/v4/components/search-input

testsuite/ui/widgets/search.py
mganisin: I created new SearchInput widget following model of widgetastic that couples required features of PF4 search-input

JaurbanRH (approved):

dhlavac (approved):

PR #706, opened 2023-07-11
Test for email payment notification

dhlavac:


mganisin (approved):

JaurbanRH (approved):

testsuite/tests/ui/billing/braintree/test_invoice_payment.py
```diff
@@ -31,3 +38,13 @@ def test_api(braintree, invoice):
     """Tests basic billing scenario for Braintree gateway where billing is triggered via UI"""
     charged = invoice.charge()
     braintree.assert_payment(charged)
+
+
+def test_mail_completed_payment(invoice, mailhog_client):
```
mganisin, line 43: Can this be done without credit card?
This specific test has now two requirements 1. mailhog, 2. braintree So far both of these are either available or both unavailable. However if there will be an environment with mailhog (or mail in general) and without braintree, this still will be blocked and excluded on such environment.

dhlavac: Yes, it can, but IMHO this will be another billing scenario. It would be nice to also extend somewhen scenarios with rejected and retried payments.

PR #707, opened 2023-07-13
Added wait time for Auth0 login

jsmolar:
Added simple wait for auth0 login for X/P/Z testing

### DIFF
--- a/testsuite/ui/views/auth.py
+++ b/testsuite/ui/views/auth.py
@@ -20,7 +20,7 @@ class Auth0View(View, Navigable):
 
     def login(self, email, password):
         """Login to 3scale via Auth0"""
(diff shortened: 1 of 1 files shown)


testsuite/ui/views/auth.py
```diff
@@ -20,7 +20,7 @@ class Auth0View(View, Navigable):
 
     def login(self, email, password):
         """Login to 3scale via Auth0"""
-        self.last_login_button.wait_displayed()
+        self.last_login_button.wait_displayed(timeout="20s")
```
mganisin, line 23: Add comment please why it is needed. One day someone else will be looking at code, without context might be convinced it is unnecessary delay.

PR #708, opened 2023-07-13
Tests for reset password mails

dhlavac:


testsuite/tests/ui/mail_and_messages/test_account_messages.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Test:
+        - Create new provider account
+        - Send reset password
+        - Assert that mail with password recovery was received
+    """
+    mail = f"{provider_account_user.entity_name}@example.com"
+
+    navigator.open(LoginView, url=settings["threescale"]["admin"]["url"])
+    reset_view = navigator.navigate(ResetAdminPasswordView)
```
JaurbanRH, line 17: Why you need to open LoginView?

PR #714, opened 2023-07-25
Extend password reset tests

JaurbanRH:


testsuite/tests/ui/test_reset_password.py
```diff
@@ -0,0 +1,39 @@
...
+    provider_account_user.activate()
+    mail = f"{account_name}@example.com"
+
+    navigator.open(LoginView, url=settings["threescale"]["admin"]["url"])
+    reset_view = navigator.navigate(ResetAdminPasswordView)
+    reset_view.reset_password(mail)
+
+    message = mailhog_client.find_message(subject="Password Recovery", receiver=mail)
+
+    reset_link = re.findall("https://.*\\r", message["items"][0]["Content"]["Body"])[0][:-1]
```
dhlavac, line 29: when you use `re.search(r"(?P<url>https?://[^\s]+)", message["items"][0]["Content"]["Body"]).group("url")`  you will have only url (not list) without need to remove characters from it

testsuite/tests/ui/test_reset_password.py
```diff
@@ -0,0 +1,39 @@
+"""Test for reset password"""
```
dhlavac, line 1: Imho it can be added as an extension to the current test, no need to create a separate test for this.

testsuite/tests/ui/test_reset_password.py
```diff
@@ -0,0 +1,39 @@
...
+
+    navigator.open(LoginView, url=settings["threescale"]["admin"]["url"])
+    reset_view = navigator.navigate(ResetAdminPasswordView)
+    reset_view.reset_password(mail)
+
+    message = mailhog_client.find_message(subject="Password Recovery", receiver=mail)
+
+    reset_link = re.findall("https://.*\\r", message["items"][0]["Content"]["Body"])[0][:-1]
+    page = navigator.open(ResetPasswordView, url=reset_link, exact=True)
+    page.change_password("password")
```
dhlavac, line 31: Maybe some generated password? (it can run on public instances)

dhlavac (changes requested):

mganisin (approved):

jsmolar (approved):

dhlavac (approved):

PR #733, opened 2023-08-16
Added new account creation for account messages test

jsmolar:


testsuite/tests/ui/mail_and_messages/test_account_messages.py
```diff
@@ -1,12 +1,33 @@
...
+
+
+@pytest.fixture(scope="module")
+def account(custom_account, request, account_password):
+    """
+    Account scoped for module (rather than session) because account password is updated in this module.
+    """
+    iname = blame(request, "id")
+    account = rawobj.Account(org_name=iname, monthly_billing_enabled=False, monthly_charging_enabled=False)
+    account.update(
```
dhlavac, line 22: Maybe better to create an Account with params than update it.

dhlavac (changes requested):

testsuite/tests/ui/mail_and_messages/test_account_messages.py
jsmolar: This is how we use it in main conftest

JaurbanRH (approved):

dhlavac (approved):

PR #752, opened 2023-10-09
Add test for method/metrics deletion used by mapping rule

JaurbanRH:


testsuite/tests/system/mapping/test_delete_mapping_rules.py
```diff
@@ -0,0 +1,83 @@
...
+    metric = service.metrics.create(rawobj.Metric("metric"))
+    mapping = service.mapping_rules.create(rawobj.Mapping(metric, "/anything"))
+    return metric, mapping
+
+
+@pytest.fixture(scope="module")
+def backend(service, threescale):
+    """Gets backend used by default service"""
+    backend_id = service.backend_usages.list()[0]["backend_id"]
+    return threescale.backends.get(backend_id)
```
mkudlej, line 37: prefer to use `read` method instead of `get`

mkudlej (commented):
LGTM except of one small comment

mkudlej (approved):

mdujava (approved):

PR #794, opened 2024-02-07
Fixed admin users table locator

dhlavac:


mdujava (approved):

mijaros (commented):
Wouldn't it make sense to use constant, that would contain the location of the column? It seems to be the same for both and it's always better to use some named symbol instead of magical constants.

mijaros (approved):

mdujava (approved):

PR #798, opened 2024-02-07
Suggestion - Make default python version 3.11

dhlavac:


Pipfile
```diff
@@ -64,4 +64,4 @@ pytest-html = "*"
 ansi2html = "*"
 
 [requires]
-python_version = "3"
+python_version = "3.11"
```
pehala, line 67: This will not allow 3.12 or other newer versions in future

mkudlej (changes requested):
I agree with Peter. Let's leave it as it is.

PR #808, opened 2024-02-26
Make TLS tests compatible with pytest-persistence plugin 

JaurbanRH:


mdujava (approved):

dhlavac:
@JaurbanRH there are still some tests marked as no persistence in the tls tests folder, is it correct ? Or should we remove marks from that tests ?

mdujava (changes requested):
`resultsdir` should be set by Makefile, but running `make` without setting `resultsdir` explicitly, certs are stored in `/tmp`; maybe `resultsdir` is not exported so pytests does not see it?

JaurbanRH:
> @JaurbanRH there are still some tests marked as no persistence in the tls tests folder, is it correct ? Or should we remove marks from that tests ?

@dhlavac Yes that is correct. Those test need complete rewrite to be compatible with persistence plugin.

mdujava (approved):

PR #831, opened 2024-05-07
Permissions tests

dhlavac:
Base setup of permissions tests, more are coming.

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -14,6 +14,7 @@ import backoff
 import importlib_resources as resources
 import openshift_client as oc
 import pytest
+from typing import List
 from dynaconf.vendor.box.exceptions import BoxKeyError
 from pytest_metadata.plugin import metadata_key
 from threescale_api import client, errors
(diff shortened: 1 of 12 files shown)


testsuite/tests/conftest.py
```diff
@@ -529,6 +530,31 @@ def _custom_user(params, autoclean=True):
...
+        username = blame(request, "pa")
+        user_params = rawobj.AccountUser(username=username, email=f"{username}@example.com", password=account_password)
+        user = threescale.provider_account_users.create(params=user_params)
+        user.set_role_member()
+        user.activate()
+        user.permissions_update(allowed_services=allowed_services, allowed_sections=allowed_sections)
+
+        if autoclean and not testconfig["skip_cleanup"]:
+            request.addfinalizer(user.delete)
+        return user
```
JaurbanRH, line 553: Please move deletion of user right after creation (line 547).

testsuite/tests/ui/users_and_roles/test_permissions.py
```diff
@@ -0,0 +1,83 @@
...
+        BillingSettingsView,
+        marks=[pytest.mark.xfail, pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-10995")],
+    ),
+    pytest.param("plans", ActiveDocsView),
+]
+
+
+def get_views_by_permission(permission_name):
+    """Returns a list of views that has the given permission by permission name"""
+    return [view for perm, view in PERMISSION_DICT if perm == permission_name]
```
mdujava, line 32: No usage for this function, and I believe `perm, view in PERMISSION_DICT` would result in too many values to unpack.

mdujava (approved):

JaurbanRH (approved):

PR #836, opened 2024-05-20
Add way to configure rhsso/rhbk via settings

mdujava:
- onboards on 3scale-api 0.35.0
- allows to specify rhsso.kind to be rhsso or rhbk
- fix problem with not respecting fixtures.tools.namespace for sso lookup
- add --sso-only flag to skip non sso tests


testsuite/dynaconf_loader.py
```diff
@@ -109,24 +109,33 @@ def _apicast_image(ocp):
...
+                return tools.secrets["credential-sso"]["ADMIN_PASSWORD"].decode("utf-8")
         except (OpenShiftPythonException, KeyError):
             try:
                 # try 7.4 known deployment if previous fails
                 return tools.environ("dc/sso")["SSO_ADMIN_PASSWORD"]
             except (OpenShiftPythonException, KeyError):
-                return None
+                pass
+
+    return None
```
mijaros, line 138: This depth seems kinda redundant and deep, every positive branch ends with the return so nesting the code into the except branches is not good idea to be honest.

testsuite/tests/conftest.py
```diff
@@ -183,6 +184,27 @@ def pytest_collection_modifyitems(session, config, items):
...
+        to_remove = True
+        for fixture in selected_fixtures:
+            if fixture in item.fixturenames:
+                to_remove = False
+
+        if to_remove:
+            selected.remove(item)
+            deselected.append(item)
+
+    items[:] = selected
```
mijaros, line 205: I want on the record that I don't like this! :-)

mijaros (changes requested):
I think that the changes are well written and i understand that the nesting is relict of older development, but why don't fix it now?

mijaros (approved):
Ok, this looks very promissing.

testsuite/tests/conftest.py
dhlavac, line 205: Cannot be simplified to ?
`   selected_fixtures = {"rhsso_setup"}
    deselected = []

    sso_only_opt = config.option.sso_only
    selected = []

    for item in items:
        for marker in item.iter_markers(name="issue"):
            issue = marker.args[0]
            issue_id = issue.rstrip("/").split("/")[-1]
            item.user_properties.append(("issue", issue))
            item.user_properties.append(("issue-id", issue_id))

        if sso_only_opt and not any(fixture in selected_fixtures for fixture in item.fixturenames):
            deselected.append(item)
        else:
            selected.append(item)

    items[:] = selected

    if deselected:
        config.hook.pytest_deselected(items=deselected)`

JaurbanRH (dismissed):

Pipfile
```diff
@@ -60,6 +60,8 @@ pytest-persistence = "*"
 pytest-metadata = ">=3.0.0"
 pytest-html = "*"
 ansi2html = "*"
+ipdb = "*"
```
mkudlej, line 63: This should e part of different PR.

mkudlej (approved):
LGTM except of one change in Pipfile, which should be removed I think.

mkudlej (approved):

testsuite/tests/conftest.py
```diff
@@ -176,13 +177,28 @@ def pytest_collection_modifyitems(session, config, items):
...
+    deselected = []
+
     for item in items:
         for marker in item.iter_markers(name="issue"):
             issue = marker.args[0]
             issue_id = issue.rstrip("/").split("/")[-1]
             item.user_properties.append(("issue", issue))
             item.user_properties.append(("issue-id", issue_id))
 
+        if sso_only_opt and not any(fixture in selected_fixtures for fixture in item.fixturenames):
```
dhlavac, line 193: we need to deselect performance tests from sso tag, there are 4 tests from performance

dhlavac (approved):

PR #838, opened 2024-05-23
Save pytest_metadata as json file

mastastny:
Saves metadata only when pytest is executed with --tool-check

### DIFF
--- a/testsuite/tests/tools/conftest.py
+++ b/testsuite/tests/tools/conftest.py
@@ -1,0 +1,38 @@
+"""Tools conftest"""
+
+import json
+import os
+from pathlib import Path
+import pytest
+
+
+def get_results_dir_path():
+    """Method that gives you the path to the directory where you should store test results."""
+    try:
+        results_dir = os.environ.get("resultsdir")
+        return Path(results_dir)
+    except KeyError:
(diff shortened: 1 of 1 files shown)


testsuite/tests/tools/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Save data to the attachments directory in json format.
+    If the attachments directory does not exist create it.
+    Args:
+         :param data: Data to be stored
+         :param f_name: name of the created file
+    """
+    current_file = Path(__file__).resolve()
+    project_dir = current_file.parents[3]
+    attachments_dir = project_dir / "attachments"
```
mdujava, line 18: I would use something like https://github.com/3scale-qe/3scale-tests/blob/9a92b301043622902bcfff3b6f678a486dae46df/testsuite/tests/ui/conftest.py#L437 honoring resultsdir env variable.

mastastny: Done

JaurbanRH (approved):

mdujava (dismissed):

mdujava (approved):

JaurbanRH (approved):

PR #876, opened 2024-09-25
Fix pylint error rename

mdujava:
Also removing container runtimes as they are not used, and it looks like podman did change the Image class member "container", which most probably broke `PodmanRuntime`.


mastastny (approved):

dhlavac:
@mdujava maybe ignore `too-many-positional-arguments`  instead of adding ignores

dhlavac (approved):

PR #878, opened 2024-09-30
Make toolbox cli tests run on other workers

mdujava:
Mark all non-cli tests with a group, so they run sequentially. Speed up from ~29min to ~24min for `make toolbox`.


mkudlej (approved):
LGTM

mkudlej (approved):

PR #893, opened 2024-11-22
Fix test metrics hierarchies

mastastny:
Error was caused by deletion of custom_ui_backend while it was still registered in custom_ui_product.
This fix assures, that teardown of custom_ui_product fixtures will be executed before teardown of custom_ui_backend.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -9,7 +9,6 @@ types-braintree = "*"
 types-stripe = "*"
 types-Pillow = "*"
 black = "*"
-types-retry = "*"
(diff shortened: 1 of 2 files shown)


dhlavac:
Adding dependency for this teardown is kind of no go, I would suggest using one of below approaches:

- change finalizer - decorate it with backoff, create custom finalizer
- create custom_ui_service that binds by default product and backend but is parametrized and in case of this test it will be created without bind, but finalizer will handle that
- extend the test with not only adding but also unassigning backend from product

dhlavac (approved):

PR #894, opened 2024-11-25
Get version of 3scale from APIManager annotation

mdujava:


mdujava:
This is candidate to backport if 3scale backports annotations of the APIManager object

dhlavac (approved):

PR #913, opened 2025-01-16
move backend usage cleanup from backend deleting fixture to service f…

mkudlej:
…ixture + required fixes in tests

### DIFF
--- a/testsuite/tests/conftest.py
+++ b/testsuite/tests/conftest.py
@@ -717,7 +717,10 @@ def httpx():
 @pytest.fixture(scope="module")
 def service(backends_mapping, custom_service, service_settings, service_proxy_settings, lifecycle_hooks):
     "Preconfigured service with backend defined existing over whole testsing session"
-    return custom_service(service_settings, service_proxy_settings, backends_mapping, hooks=lifecycle_hooks)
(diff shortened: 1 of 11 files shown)


mdujava:
Is there a reason why those deletes are not after check for skip_clenup?

mkudlej:
+ check testsuite/tests/apicast/policy/rate_limit_headers/test_rate_limit_headers.py - why there is rename to service2

testsuite/tests/conftest.py
```diff
@@ -930,7 +933,8 @@ def finalizer():
                             hook(svc)
                         except Exception:  # pylint: disable=broad-except
                             pass
-
+                    for usage in svc.backend_usages.list():
```
mdujava, line 936: This is reached after backend_usages are deleted, unless order of fixtures is changed in a tests to have `backend_usages` before `service`.

mdujava:
in each of 
```
testsuite/tests/apicast/policy/keycloak_role_check/test_keycloak_policy.py
testsuite/tests/apicast/policy/keycloak_role_check/test_keycloak_policy_combined.py
```
last test executed will hit ERROR: `threescale_api.errors.ApiClientError: Response(403 Forbidden): b'{"errors":{"base":["cannot be deleted because it is used by at least one Product"]}}'`

PR #920, opened 2025-03-21
add tests for conditional policy and policies, which can be used with it

mkudlej:
Detail see here https://issues.redhat.com/browse/THREESCALE-9320

### DIFF
--- a/testsuite/tests/apicast/policy/conditional/test_batcher_policy_mapping_rules.py
+++ b/testsuite/tests/apicast/policy/conditional/test_batcher_policy_mapping_rules.py
@@ -1,0 +1,101 @@
+"""
+Test checks that batcher policy works with conditional policy.
+"""
+
+from time import sleep
+import pytest
+
+from packaging.version import Version  # noqa # pylint: disable=unused-import
+from testsuite import TESTED_VERSION, rawobj  # noqa # pylint: disable=unused-import
+
+
+pytestmark = [
+    pytest.mark.nopersistence,
+    pytest.mark.skipif("TESTED_VERSION < Version('2.16')"),
+]
+
+BATCH_REPORT_SECONDS = 50
+
+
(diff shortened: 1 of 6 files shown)


testsuite/tests/apicast/policy/conditional/test_batcher_policy_mapping_rules.py
```diff
@@ -0,0 +1,97 @@
...
+    """Set policy settings"""
+    return rawobj.PolicyConfig(
+        "conditional",
+        {
+            "condition": {
+                "operations": [
+                    {"left": "{{ uri }}", "left_type": "liquid", "op": "==", "right": "/get1", "right_type": "plain"}
+                ]
+            },
+            "policy_chain": [{"name": "3scale_batcher", "configuration": {"batch_report_seconds": 50}}],
```
mdujava, line 29: I would use global constant in the testfile for batch_report_seconds, which is reused in api_client and test.

testsuite/tests/apicast/policy/conditional/test_on_failed.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Default backend with url from private_base_url.
+    """
+    return custom_backend("backend_default", endpoint=private_base_url("httpbin"))
+
+
+@pytest.fixture(scope="module")
+def policy_settings():
+    "config of upstream connection policy with read_timeout set"
+    return [
```
mdujava, line 23: in other tests policy_settings is not wrapped in a list.

mdujava (changes requested):
Also on alpha build there was one fail: 
```
FAILED testsuite/tests/apicast/policy/conditional/test_retry.py::test_retry - AssertionError: assert 500 == 200
```

mdujava (approved):

mastastny (approved):

PR #927, opened 2025-04-30
Fapi advanced

mastastny:


testsuite/rhsso/__init__.py
```diff
@@ -26,13 +27,13 @@ def __init__(self, rhsso: RHSSO, realm: Realm, client: Client, user, username, p
...
         self.password = password
-        self._oidc_client = None
+        self._oidc_client: KeycloakOpenID | None = None
 
     @property
     def oidc_client(self) -> KeycloakOpenID:
         """OIDCClient for the created client"""
         if not self._oidc_client:
-            self._oidc_client = self.client.oidc_client
+            self._oidc_client = cast(KeycloakOpenID, self.client.oidc_client)
```
mdujava, line 36: cast is not needed here , I belive typing the oidc_client property in Client class will fix any errors here.

mdujava: After https://github.com/marcospereirampj/python-keycloak/pull/655 will be merged this will work: 
```patch
diff --git a/testsuite/rhsso/__init__.py b/testsuite/rhsso/__init__.py
index e34e1f00..8c01ddf4 100644
--- a/testsuite/rhsso/__init__.py
+++ b/testsuite/rhsso/__init__.py
@@ -26,7 +26,7 @@ class RHSSOServiceConfiguration:
         self.client = client
         self.username = username
         self.password = password
-        self._oidc_client = None
+        self._oidc_client: KeycloakOpenID | None = None
 
     @property
     def oidc_client(self) -> KeycloakOpenID:
diff --git a/testsuite/rhsso/objects.py b/testsuite/rhsso/objects.py
index 88ba962c..e678705c 100644
--- a/testsuite/rhsso/objects.py
+++ b/testsuite/rhsso/objects.py
@@ -42,7 +42,7 @@ class Realm:
         self.admin.update_user(user_id, {"emailVerified": True})
         return user_id
 
-    def oidc_client(self, client_id, client_secret):
+    def oidc_client(self, client_id, client_secret) -> KeycloakOpenID:
         """Create OIDC client for this realm"""
         return KeycloakOpenID(
             server_url=self.admin.connection.server_url,
@@ -68,7 +68,7 @@ class Client:
         self.admin.assign_client_role(user["id"], realm_management, role)
 
     @property
-    def oidc_client(self):
+    def oidc_client(self) -> KeycloakOpenID:
         """OIDC client"""
         # Note This is different clientId (clientId) than self.client_id (Id), because RHSSO
         client_id = self.admin.get_client(self.client_id)["clientId"]

```

mdujava (changes requested):
Few comments from development needs to be removed.
Also typing 2 functions relates to bumping the keycloak library that can be done in separate PR #936, so that commit can be left out from this PR

testsuite/rhsso/__init__.py
mastastny: Fixed

mdujava (changes requested):

testsuite/rhsso/objects.py
```diff
@@ -42,23 +42,26 @@ def create_user(self, username, password, **kwargs):
...
-    def oidc_client(self, client_id, client_secret) -> KeycloakOpenID:
+    def oidc_client(self, client_id, client_secret, cert=None) -> KeycloakOpenID:
         """Create OIDC client for this realm"""
         return KeycloakOpenID(
             server_url=self.admin.connection.server_url,
             client_id=client_id,
             realm_name=self.name,
             client_secret_key=client_secret,
+            cert=cert,
+            verify=False,
```
mdujava, line 53: This might need to be configurable in the future, if the SSO will be using same cert as cluster, it might make sense to verify.

mdujava (changes requested):

mdujava (dismissed):

testsuite/tests/conftest.py
```diff
@@ -627,7 +627,9 @@ def _resolve_rhsso(testconfig, tools, rhsso_kind):
     key = "no-ssl-rhbk"
```
mdujava, line 627: it's selecting no-ssl-rhbk, is there any discovery for ssl variant, and skip if such rhbk is not found?

mdujava (dismissed):

cathal-bailey (dismissed):

mastastny:
> @mastastny any update there ?

@dhlavac waiting for review

cathal-bailey (dismissed):

testsuite/rhsso/objects.py
```diff
@@ -7,29 +7,41 @@
 from testsuite.config import settings
 
 
+def _fallback(value, default):
+    return default if value is None else value
```
mdujava: not sure if it should be in utils, and not here

mastastny: moved

testsuite/rhsso/objects.py
```diff
@@ -5,31 +5,40 @@
...
 from testsuite.config import settings
+from testsuite.utils import fallback
 
 
 class Realm:
     """Helper class for RHSSO realm manipulation"""
 
-    def __init__(self, master: KeycloakAdmin, name) -> None:
+    def __init__(self, master: KeycloakAdmin, name, verify=None) -> None:
+        self.verify = verify if verify else settings["ssl_verify"]
```
mdujava: if verify is false settings should be used instead?
if not, fallback can be used

mdujava (changes requested):

mdujava (approved):

cathal-bailey (approved):

PR #936, opened 2025-05-21
Bump python-keycloak library

mastastny:


mdujava (dismissed):

mdujava (dismissed):

testsuite/rhsso/objects.py
```diff
@@ -101,6 +102,7 @@ def __init__(self, server_url, username, password) -> None:
                 realm_name="master",
                 verify=False,
             )
+            self.master.get_clients()  # test whether the server url is valid
```
dhlavac, line 105: maybe add  catch with log message if it fails in the end ?

mdujava: Yeah, I would keep behavior as is. If we might want to change it in the future for example: add Capability.SSO if sso was discovered and functional else skipping the tests. I would suggest to get it in as a separate MR.

mdujava (approved):

mdujava (approved):

PR #937, opened 2025-05-22
deselect tests instead of skipping them in runtime

mdujava:
Number of tests are identical when running these make targets:
check: 11
smoke: 10
speedrun: 339
sandbag: 204
disruptive: 274
flaky: 363
ui: 216
toolbox: 143
fuzz: 120


dhlavac:
Yes, finally !!!! This was RFE for a long time !

PR #976, opened 2026-01-07
Fix mypy type error in RHSSO objects

cathal-bailey:
Add type guard to ensure server_url is non-None string before passing to KeycloakOpenID constructor. Fixes argument type incompatibility error.


mdujava:
related change in a keycloak lib: https://github.com/marcospereirampj/python-keycloak/pull/687

mdujava (approved):
this is fine, but if library fixes `def server_url` to return just str, we should remove this

PR #982, opened 2026-01-29
add pip-system-certs, that will automaticly use system tls ca

mdujava:


mdujava:
another possible solution is to set `SSL_CERT_FILE` like in Makefile either to system ca, or create ca-bundle on fly.

Container builds use `SSL_CERT_FILE`.

azgabur (approved):
What problem is this change trying to solve?

mdujava:
If you have `verify=True` and cluster with cert from custom ca (installed in system with update-ca-trust), you will see error unknown CA (rhbk route to be exact)

dhlavac (approved):

PR #1016, opened 2026-03-11
Adding waits to resolve timing issues on test_devel_sections.py

cathal-bailey:
- Tests were failing ~50/60% of the time due to race condition/timing issue in dev portal
- Using @backoff to retry on failed assertions. Pytestmark added to facilitate backoff decorator. 
- Reliability of test greatly improved with this change. No fails observed locally after addition of wait times.

### DIFF
--- a/testsuite/tests/ui/devel/test_devel_sections.py
+++ b/testsuite/tests/ui/devel/test_devel_sections.py
@@ -1,5 +1,6 @@
 """Test for developer portal sections"""
 
+import backoff
 import pytest
 from packaging.version import Version  # noqa # pylint: disable=unused-import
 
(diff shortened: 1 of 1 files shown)


testsuite/tests/ui/devel/test_devel_sections.py
```diff
@@ -85,11 +90,18 @@ def test_dev_portal_sections(account, custom_devel_login, browser, testconfig, d
...
+    # The group was assigned in fixture, but permissions need time to sync to portal
+    time.sleep(2)
+
     browser.url = testconfig["threescale"]["devel"]["url"] + dev_portal_page
 
     assert browser.element(".//h1").accessible_name == "Test"
 
     custom_devel_login(name="john", password="123456", fresh=True)
+
+    time.sleep(2)
```
mdujava: are those permission updated when user is logging in? otherwise I don't think this sleep is needed.

mastastny (commented):
We don't use ```sleep``` unless it is the only option. In this case, I would rather use ```backof``` library, as we do in other tests.

mdujava (changes requested):
This patch contains unnecessary move of pytest.makr.issue and new blank lines

cathal-bailey:
> This patch contains unnecessary move of pytest.makr.issue and new blank lines

When I was testing, there appeared to be a conflict between the `@backoff` and `@pytest.mark.skipif("TESTED_VERSION < Version('2.14-dev')")`. I moved these decorators to module-level `pytestmark` block and this resolved the errors I was seeing. 

I will remove the unnecessary blank lines and re-commit, and I can also dig deeper into the errors coming from TESTED_VERSION, if you would rather leave the decorators as they were.

mdujava (approved):

mdujava (approved):

mastastny (approved):

PR #1017, opened 2026-03-13
Adding explicit call for apicast-production metrics

cathal-bailey:
- The test queries Prometheus for metrics for "apicast-staging" and "apicast-production"
- Since production has never received HTTP traffic, request-dependent metrics don't exist and the test fails 
- For production metrics, promotes staging config to production and sends a GET request to /get endpoint 
- This generates the missing metrics (apicast_status, upstream_status, threescale_backend_calls, and response time histograms - **5 tests now passing**)
- prometheus.wait_on_next_scrape so that metrics appear in db, as seen elsewhere for prometheus/apicast in testsuite
- All metrics passing on Alpha with this change, but considerable increase in test time (now ~4 mins)
- Marking apicast-prodiuction as disruptive to ensure apicast restart does not impact other tests. 
- Adding idempotent check to prod_client fixture to prevent duplicate promotion errors

### DIFF
--- a/testsuite/tests/prometheus/apicast/test_metrics.py
+++ b/testsuite/tests/prometheus/apicast/test_metrics.py
@@ -40,9 +40,36 @@ STATUSES = [300, 418, 507]
 
 
 # pylint: disable=unused-argument
-@pytest.fixture(scope="module", params=["apicast-staging", "apicast-production"])
-def metrics(request, prometheus):
+@pytest.fixture(
+    scope="module", params=["apicast-staging", pytest.param("apicast-production", marks=pytest.mark.disruptive)]
+)
+def metrics(request, prometheus, application, api_client):
     """Return all metrics from target defined of staging and also production apicast."""
+    # Check if any required metrics don't exist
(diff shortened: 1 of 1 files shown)


testsuite/tests/prometheus/apicast/test_metrics.py
```diff
@@ -41,8 +41,35 @@
...
+        for metric in required_histogram_metrics
+        for suffix in ["_bucket", "_sum", "_count"]
+    )
+
+    # If some metrics do not exist, trigger with explicit HTTP request
+    if standard_missing or histogram_missing:
+        if request.param == "apicast-production":
+            service = application.service
+            version = service.proxy.list().configs.latest()["version"]
+            service.proxy.list().promote(version=version)
```
mdujava: not liking this code reminded me that `.promote` on a apicast-production will cause restart of apicast, which can affect running tests in parallel, so apicast-production would need to set pytest.mark.disruptive

mdujava (dismissed):

mastastny (dismissed):

PR #1021, opened 2026-03-23
Fix billing tests to pass if default provider id != 2

mdujava:


cathal-bailey (dismissed):

mastastny (changes requested):
In the test_mail_completed_payment there is subject hardcoded to default tenant with name "Provider Name", so if we use private_tenant, the test fails. I would change it to:
```subject=f"{provider_account['org_name']} API - Payment completed"``` 

Also for the private_tenant the normalize_url function doesn't work well and the replacement should be changed from this:```("3scale-admin", "3scale")``` to this: ```("-admin.", ".")```.

mastastny (dismissed):

mdujava:
when resolving conflict I wrongly reverted one function `R6301: Method could be a function (no-self-use)` caught it.

mastastny (approved):

mastastny (approved):

cathal-bailey (approved):

PR #1022, opened 2026-03-30
Add backoff retry for RHSSO and Content Limits tests 

cathal-bailey:
- Add `@backoff` decorator to `staging_client` fixture
  - RHSSO tests have been failing frequently on Jenkins pipeline runs for Alpha build
  - Same fix added for `payload_limits`, flaky on Jenkins pipeline
  - The tests will now retry on 404 status (config not loaded yet)
  - Should fix timing issues in CI/Jenkins where APIcast staging hasn't polled configuration yet

### DIFF
--- a/testsuite/tests/apicast/auth/rhsso/openid_rhsso/conftest.py
+++ b/testsuite/tests/apicast/auth/rhsso/openid_rhsso/conftest.py
@@ -2,6 +2,7 @@
 Conftest for the openid rhsso credentials locations tests
 """
 
+import backoff
 import pytest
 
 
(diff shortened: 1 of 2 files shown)


testsuite/tests/apicast/auth/rhsso/openid_rhsso/conftest.py
```diff
@@ -14,6 +15,21 @@ def staging_client(api_client):
     """
     client = api_client()
```
mdujava, line 16: There should be default retry on 503 and 404 
https://github.com/3scale-qe/3scale-api-python/blob/32de76d9eee8fee81ec3ed9876eaea7d4806338d/threescale_api/utils.py#L56

maybe better is to parametrize this fixture or just create api_client in a tests so auth is not set to None, and then internal retry should work

cathal-bailey:
Opened #1024

PR #1024, opened 2026-04-17
Change to resolve 404s seen in RHSSO tests on Jenkins pipeline

cathal-bailey:
- Refactoring #1022 (now closed)
- Custom `@backoff` removed from RHSSO tests in favour of built-in retries for 404/503s
- `api_client` and `prod_client` are module scoped, should prevent double promotion in production 
- Explicit `client.auth = None` for each test. Will need to validate built-in retry using Jenkins pipeline.


mdujava:
I do not believe that builtin retries will work as expected:
* either in place of calling .get auth is None so Session could not be created (session is required for retry logic)
* or api_client has cached Session with default auth (before assigning None to it in tests)

mdujava:
What is a status of this? should we close this?

PR #1029, opened 2026-05-01
Test Updates to Align with Header Removal from APIcast Midstream

cathal-bailey:
- APIcast midstream update removes the Server header from HTTP responses. This breaks tests that rely on checking headers to confirm APIcast is responding.
- Update to the backoff mechanism to look for valid response code rather than server header. Ensures backoff logic remains in place and working, in case it is needed. 
- `test_do_not_send_openresty_version.py‎` also affected by midstream change. 
- Test now verifies that no server header or version information is disclosed anywhere.

### DIFF
--- a/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
+++ b/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
@@ -54,7 +54,7 @@ def status_code(chain_name, on_failed_configuration) -> int:
 
 @backoff.on_predicate(
     backoff.fibo,
-    lambda response: response.headers.get("server") not in ("openresty", "envoy"),
+    lambda response: response.status_code not in (200, 503, 500),
     max_tries=8,
     jitter=None,
 )
@@ -72,4 +72,3 @@ def test_on_failed_policy(application, status_code):
 
     response = make_request(api_client)
     assert response.status_code == status_code
-    assert response.headers["server"] in ("openresty", "envoy")
--- a/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
+++ b/testsuite/tests/apicast/policy/on_failed/test_onfailed.py
@@ -52,15 +52,16 @@ def status_code(chain_name, on_failed_configuration) -> int:
     return on_failed_configuration.get("error_status_code", 503)
 
 
-@backoff.on_predicate(
-    backoff.fibo,
-    lambda response: response.status_code not in (200, 503, 500),
-    max_tries=8,
-    jitter=None,
-)
-def make_request(api_client):
+def make_request(api_client, expected_code):
     """Make request to the product and retry if the response isn't from APIcast"""
-    return api_client.get("/")
+
+    @backoff.on_predicate(
+        backoff.fibo, lambda response: response.status_code != expected_code, max_tries=8, jitter=None
+    )
+    def _request():
+        return api_client.get("/")
+
+    return _request()
 
 
 def test_on_failed_policy(application, status_code):
@@ -70,5 +71,5 @@ def test_on_failed_policy(application, status_code):
     """
     api_client = application.api_client(disable_retry_status_list=(503,))
 
-    response = make_request(api_client)
+    response = make_request(api_client, status_code)
     assert response.status_code == status_code
--- a/testsuite/tests/apicast/test_do_not_send_openresty_version.py
+++ b/testsuite/tests/apicast/test_do_not_send_openresty_version.py
@@ -29,9 +29,7 @@ def client(api_client):
     return api_client(disable_retry_status_list={503, 404})
 
 
-@backoff.on_predicate(
-    backoff.fibo, lambda x: x.headers.get("server", "") not in ("openresty", "envoy"), max_tries=8, jitter=None
-)
+@backoff.on_predicate(backoff.fibo, lambda x: x.status_code != 503, max_tries=8, jitter=None)
 def make_requests(client):
     """Make sure that we get 503 apicast (backend is not available)"""
     return client.get("/anything")
@@ -41,15 +39,14 @@ def make_requests(client):
 def test_do_not_send_openresty_version(client):
     """
     Make request to non existing endpoint
-    Assert that the response does not contain openresty version in the headers
+    Assert that any server header (including envoy) does not contain "/", indicating version
     Assert that the response does not contain openresty version in the body
     """
     response = make_requests(client)
     assert response.status_code == 503
 
-    assert "server" in response.headers
-    if response.headers["server"] == "envoy":
-        pytest.skip("envoy edge proxy in use")
-    assert response.headers["server"] == "openresty"
+    server_header = response.headers.get("server", "")
+    if server_header:
+        assert "/" not in server_header
 
-    assert "<center>openresty</center>" in response.text
+    assert "openresty/" not in response.text


mdujava (dismissed):

mdujava:
are there any upstream changes or jira issue that shows this is expected change?

cathal-bailey:
> are there any upstream changes or jira issue that shows this is expected change?

- Yes, headers were explicitly removed here https://gitlab.cee.redhat.com/3scale/apicast-midstream/-/merge_requests/722.

- https://redhat.atlassian.net/browse/THREESCALE-14703 - the relevant Jira issue

mdujava:
https://redhat.atlassian.net/browse/THREESCALE-4745

testsuite/tests/apicast/policy/on_failed/test_onfailed.py
```diff
@@ -54,7 +54,7 @@ def status_code(chain_name, on_failed_configuration) -> int:
 
 @backoff.on_predicate(
     backoff.fibo,
-    lambda response: response.headers.get("server") not in ("openresty", "envoy"),
+    lambda response: response.status_code not in (200, 503, 500),
```
mastastny: There is bug. In the test where you are waiting for the 503, it can pass when it gets 200, then the test will fail on the wrong status code, but the only issue is that the test didn't wait enough time to load the configuration.

I also don't like that the range of the status codes is hardcoded. If there isn't other way how to detect that configuration is loaded, I would at least test the backoff against fixture ```status_code``` and set it to something, that doesn't collide with standard status codes (e.g 418) .

mastastny (changes requested):

testsuite/tests/apicast/policy/on_failed/test_onfailed.py
cathal-bailey: I could set a global variable `expected_code` returned from fixture `status_code`, then retry while `response.status_code != expected_code`? Then when retry passes, the response code will be the correct one

mastastny: I'm not sure how do you want set the global variable, but test ```response.status_code``` against ```expected_code``` seems like good approach to me. I would personally refactor ```make_request``` function, so it accepts status code as parameter, wrap ```return api_client.get("/")``` by a nested function and anotate that nested function by ```backoff``` with the correct condition.

mastastny (approved):

testsuite/tests/apicast/test_do_not_send_openresty_version.py
```diff
@@ -29,9 +29,7 @@ def client(api_client):
     return api_client(disable_retry_status_list={503, 404})
 
 
-@backoff.on_predicate(
-    backoff.fibo, lambda x: x.headers.get("server", "") not in ("openresty", "envoy"), max_tries=8, jitter=None
-)
+@backoff.on_predicate(backoff.fibo, lambda x: x.status_code != 503, max_tries=8, jitter=None)
```
mdujava: 503 is retried by default see Application.api_client at https://github.com/3scale-qe/3scale-api-python/blob/11a7d1ccd7ca83f33d6b9028769a27ccb5591719/threescale_api/resources.py#L1432 and default HTTPClient at  https://github.com/3scale-qe/3scale-api-python/blob/11a7d1ccd7ca83f33d6b9028769a27ccb5591719/threescale_api/utils.py#L61

testsuite/tests/apicast/test_do_not_send_openresty_version.py
```diff
@@ -41,15 +39,14 @@ def make_requests(client):
...
-    assert "server" in response.headers
-    if response.headers["server"] == "envoy":
-        pytest.skip("envoy edge proxy in use")
-    assert response.headers["server"] == "openresty"
+    server_header = response.headers.get("server", "")
+    if server_header:
+        assert "/" not in server_header
 
-    assert "<center>openresty</center>" in response.text
+    assert "openresty/" not in response.text
```
mdujava: I think even server name `nginx` or `openresty` is omitted; '/' should not be there

testsuite/tests/apicast/test_do_not_send_openresty_version.py
cathal-bailey: this `@backoff` retry is required because the in-built retry is disabled in line 29. The test is configured to use a route that will return a 503 status code and then inspect the result. If the in-built 503 retry was used it would exhaust its retries and the test would raise an error.

mdujava: Was this backoff ever hit in your testing? 
(I had reverted understanding of this backoff) its retrying until return code is 503, what is a scenario that will cause this call to return something else than 503 at first call?

I think original backoff was retrying on 503s from openshift (waiting on apicast to start), but this change will not skip the 503 returned by openshift

cathal-bailey: The original intention of the backoff was to allow time for APIcast to be ready and to ensure we're testing the 503 response prior to the test. I just tested with a print debug statement and repeated the run 10 times, the backoff wasn't triggered and all tests passed. I'm thinking I can just remove the backoff entirely and simply things. WDYT?

mdujava: original backoff was retrying on openshift's 503; so I would keep that intent, retry untill Apicast returns.

when hit apicast as expected:
```
15:24:36 +0200 INFO:threescale_api.utils:[CLIENT]: curl -X GET -H 'User-Agent: python-requests/2.34.2' -H 'Accept-Encoding: gzip, deflate, zstd' -H 'Accept: */*' -H 'Connection: keep-alive' 'https://svc-mdujav--3scale-apicast-staging.a0.fqdn.example.com:443/anything?user_key=KEY'
15:24:36 +0200 INFO:threescale_api.utils:[CLIENT]:
HTTP/1.1 503 Service Temporarily Unavailable
date: Wed, 12 Aug 2026 13:24:36 GMT
content-type: text/html
content-length: 162
set-cookie: X=Y; path=/; HttpOnly; Secure; SameSite=None

<html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
</body>
</html>...
15:24:36 +0200 INFO:testsuite.tests.apicast.test_do_not_send_openresty_version:{'date': 'Wed, 12 Aug 2026 13:24:36 GMT', 'content-type': 'text/html', 'content-length': '162', 'set-cookie': 'X=Y; path=/; HttpOnly; Secure; SameSite=None'}
PASSED   [100%]
```

when hit openshift and not retried (false positive) :
```
15:29:50 +0200 INFO:threescale_api.utils:[CLIENT]: curl -X GET -H 'User-Agent: python-requests/2.34.2' -H 'Accept-Encoding: gzip, deflate, zstd' -H 'Accept: */*' -H 'Connection: keep-alive' 'https://svc-mdujav--3scale-apicast-staging.a0.fqdn.example.com:443/anything?user_key=KEY'
15:29:50 +0200 INFO:threescale_api.utils:[CLIENT]:
HTTP/1.1 503 Service Unavailable
pragma: no-cache
cache-control: private, max-age=0, no-cache, no-store
content-type: text/html

<html>
  <head>
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <style type="text/css">
      body {
        font-family: "H...
15:29:50 +0200 INFO:testsuite.tests.apicast.test_do_not_send_openresty_version:{'pragma': 'no-cache', 'cache-control': 'private, max-age=0, no-cache, no-store', 'content-type': 'text/html'}
PASSED  [100%]
```

mastastny (approved):

mdujava (approved):

PR #1032, opened 2026-05-07
Update MailhogClient method call in test_message_counter.py

cathal-bailey:
- MailHog method update triggering fails on Alpha build
- Method name updated from `find_message()` to `find_messages()` in test_message_counter.py
- Relates to https://github.com/3scale-qe/3scale-tests/pull/1015 in which method was renamed


mdujava (approved):
it's good to include something like `Fixes: 6eaad63fb154 ("update to maintained fork of backoff")`

mastastny (approved):

PR #1033, opened 2026-05-11
Fix mypy 2.0.0 var-annotated errors in toolbox tests

mastastny:


mdujava (approved):
yeah, for not being uniform, might be same as `# noqa` :)

cathal-bailey (approved):

PR #1034, opened 2026-05-11
Add tests for the bot protection of the admin portal

mastastny:
build on top of PR #1033

### DIFF
--- a/testsuite/tests/ui/auth/test_login_recaptcha.py
+++ b/testsuite/tests/ui/auth/test_login_recaptcha.py
@@ -1,0 +1,75 @@
+"""Test for login into admin portal with bot protection (recaptcha) enabled"""
+
+import pytest
+from packaging.version import Version
+
+from testsuite import TESTED_VERSION, settings
+from testsuite.ui.views.admin.login import LoginView, RequestAdminPasswordView
+from testsuite.ui.views.admin.settings.bot_protection import AdminBotProtection
+
+pytestmark = [
+    pytest.mark.usefixtures("login"),
+    pytest.mark.usefixtures("bot_protection_setup"),
+    pytest.mark.issue("https://redhat.atlassian.net/browse/THREESCALE-765"),
+    pytest.mark.skipif(TESTED_VERSION < Version("2.16"), reason="TESTED_VERSION < Version('2.16')"),
+]
+
+
+@pytest.fixture(scope="module")
+def bot_protection_setup(navigator, browser):
+    """
+    Enables admin portal bot protection via UI,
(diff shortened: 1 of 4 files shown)


testsuite/tests/toolbox/test_activedoc.py
```diff
@@ -58,7 +58,7 @@ def parse_create_command_out(output):
 
 
 # Global variable for metrics' values to check
-out_variables = {}
+out_variables: dict = {}
```
mdujava, line 60: `dict[str,str]` would be better

mdujava (changes requested):
PR took subject from wrong commit

testsuite/tests/toolbox/test_activedoc.py
mastastny: `dict[str, str] ` would not accurately describe the contents of this dictionary. For example, on line 84 `parse_create_command_out(ret["stdout"])` returns a list, and on line 85 `threescale.active_docs[...].entity` returns an entity object — neither of these are strings.

mdujava (dismissed):

dhlavac:
@mastastny rebase pls

testsuite/tests/ui/auth/test_login_recaptcha.py
```diff
@@ -0,0 +1,70 @@
...
+    yield
+
+    browser.selenium.get(settings["threescale"]["admin"]["url"])
+    for cookie in saved_cookies:
+        browser.selenium.add_cookie(cookie)
+    browser.selenium.refresh()
+    navigator.navigate(AdminBotProtection).disable_protection()
+
+
+def test_admin_login_blocked_by_recaptcha(navigator):
```
dhlavac, line 41: not applicable to all versions of 3scale, please add for versions lower than 2.16

mastastny: fixed

mdujava: this bot protection is different new in alpha, so no need for backporting to 2.16?

mastastny: The feature has been there since 2.16.0, but we officially support it since 2.16.4, so from my point of view it can be backported, but it's not necessary.

mdujava:
@dhlavac `5 passed, 1 xfailed` LGTM for me

mdujava (dismissed):

mdujava (approved):

cathal-bailey (approved):

PR #1035, opened 2026-05-14
Testsuite Python Update

cathal-bailey:
- Updating testsuite Python version from `3.11 -> 3.14`

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -14,7 +14,7 @@ jobs:
         id: setup-python
         uses: actions/setup-python@v6
         with:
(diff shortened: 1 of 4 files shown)


mdujava:
LGTM!

mdujava:
new pipenv complains `Warning: Your Pipfile requires "python_version" 3, but you are using 3.14.0 from /home/m/.local/share/v/3/bin/python.` so maybe we can keep 3.14 in a Pipfile and bump it together with GHA and Dockerfile

mdujava (approved):

mastastny (approved):

PR #1037, opened 2026-05-19
Adding Developer Portal Scope to Tokens Tests

cathal-bailey:
- Adding Developer Portal scope test file: GET, POST, DELETE
- Adding CMS tests for all scopes.
- Updates to Management scope: DELETE test added, logic update and XFAIL removed for `test_create_invoice_line_item`.
- DELETE added for Billing and Policy Registry scopes
- Developer Portal token added to `tokens.py`


testsuite/tests/ui/tokens/test_developer_portal.py
```diff
@@ -0,0 +1,176 @@
...
+    section_id = section["id"]
+
+    response = api_client("DELETE", f"/admin/api/cms/sections/{section_id}", token)
+    if permission[0]:
+        assert response.status_code == 200
+    else:
+        assert response.status_code == 403
+
+    try:
+        threescale.policy_registry.delete(section_id)
```
mastastny: I think, it should be `threescale.cms_sections.delete(section_id)` instead

mastastny (changes requested):
### Docstrings Don't Meet CONTRIBUTING.md Requirements

I noticed the docstrings in this PR describe expected results rather than manual reproduction steps, which doesn't quite align with CONTRIBUTING.md guidelines. I think it would be helpful to update them to include explicit steps for manual reproduction. Also, I feel the `test_developer_portal.py` module docstring could be more comprehensive

### Code Duplication

I see that `test_get_cms_templates`, `test_get_cms_sections`, and `test_get_cms_files` are nearly identical, differing only in the expected status code. I think parametrizing these tests would reduce code duplication and improve maintainability. The same duplication is also in other files.

testsuite/tests/ui/tokens/test_developer_portal.py
cathal-bailey: Thank you for review, I'll update docstrings so that they're a closer match with the guidelines in CONTRIBUTING.md, and fix that typo: `threescale.cms_sections.delete(section_id)`. 
Regarding the parameterisation, I had opted not to do this since some tests would not be easily parametrised, so I chose to keep the test output consistent across the board. But I can parametrise the CMS GET trios in the 3 files where it is easy to do, I think that makes sense.

testsuite/tests/ui/tokens/test_management.py
```diff
@@ -120,10 +131,55 @@ def test_create_provider_account(request, token, api_client, permission, threesc
...
+
+@pytest.mark.parametrize("resource", ["templates", "sections", "files"])
+def test_get_cms_resource(token, api_client, resource):
+    """
+    Using a Management-scoped token, send a GET request to list a CMS resource.
+    Verify the response is 200 OK.
+    """
+
+    response = api_client("GET", f"/admin/api/cms/{resource}", token)
+    assert response.status_code == 200
```
mastastny: Are you sure, that cms should be accessible with the Management scoped token? The test is passing, but I have suspicion it shouldn't.

testsuite/tests/ui/tokens/test_management.py
```diff
@@ -120,10 +131,55 @@ def test_create_provider_account(request, token, api_client, permission, threesc
...
+
+def test_create_cms_section(token, api_client, request, permission):
+    """
+    Using a Management-scoped token, send a POST request to create a new CMS
+    section. Verify the response is 201 Created (write) or 403 Forbidden (read-only).
+    """
+    title = blame(request, "section")
+    params = {"title": title, "public": True, "partial_path": f"/{title}"}
+    response = api_client("POST", "/admin/api/cms/sections", token, json=params)
+    assert response.status_code == permission[1]
```
mastastny: Same as ```test_get_cms_resource```

testsuite/tests/ui/tokens/test_management.py
```diff
@@ -66,30 +73,32 @@ def test_get_service_top_applications(service, token, api_client):
...
-    Request to create line item should have status code 403
+    Using a Management-scoped token, send a POST request to create a line item on
+    an existing invoice. Verify the response is 201 Created (write) or 403 Forbidden (read-only).
     """
 
     name = blame(request, "item")
     params = {"invoice_id": invoice.entity_id, "name": name, "description": "description", "quantity": "1", "cost": 1}
     response = api_client("POST", f"/api/invoices/{invoice.entity_id}/line_items", token, json=params)
-    assert response.status_code == 403
+    assert response.status_code == permission[1]
```
mastastny: I don't see why the xfail mark was deleted and why invoices should be accessible with Management-scoped token.

cathal-bailey: Good catch; bug outlined in https://issues.redhat.com/browse/THREESCALE-761. I have kept `assert == 403` and marked `XFAIL`. I will check if there is a plan to close this issue going forward.

testsuite/tests/ui/tokens/test_management.py
cathal-bailey: Keeping `GET` and `POST` tests, but marking `XFAIL` due to same bug

mastastny (approved):

PR #1039, opened 2026-05-25
Add `container-image-temp` that will build and tag image for expiration

mdujava:
when building testing images and pushing them to quay, this will ensure that such temp images will not stay in quay.io registry


cathal-bailey (approved):

mdujava:
@3scale-qe/developer if you need to create temporary image that you will push to quay, use `make container-image-temp [...]` it will make sure that images will not stay there (you can still change expiration date via quay webUI)

PR #1040, opened 2026-05-26
Create product routes

mastastny:


testsuite/gateways/apicast/system.py
```diff
@@ -35,6 +38,54 @@ class SystemApicast(AbstractApicast):
...
+
+    def on_service_create(self, service: Service):
+        """Creates staging OCP route when zync is disabled"""
+        if not self._zync_disabled:
+            return
+        proxy = service.proxy.fetch()
+        self._create_route(self._route_name(service), proxy["sandbox_endpoint"], "apicast-staging")
+
+    def on_proxy_promote(self, service: Service):
+        """Creates production OCP route when zync is disabled"""
```
mdujava: what if route is already created, and service is updated and promoted (no change for route needed)

testsuite/gateways/apicast/system.py
```diff
@@ -35,6 +38,54 @@ class SystemApicast(AbstractApicast):
...
+            name, Routes.Types.EDGE, service=ocp_service, hostname=hostname, **{"insecure-policy": "Allow"}
+        )
+        self._routes.append(name)
+
+    def on_service_create(self, service: Service):
+        """Creates staging OCP route when zync is disabled"""
+        if not self._zync_disabled:
+            return
+        proxy = service.proxy.fetch()
+        self._create_route(self._route_name(service), proxy["sandbox_endpoint"], "apicast-staging")
```
mdujava: does this wait on a route creation? or we count on api_client to retry on 404/503?

testsuite/gateways/apicast/system.py
mdujava, line 20: I as imagining to have `class ZyncLessApicast(SystemApicast):` that will just define those lifecyclehooks, as right now rhsso wont work. but if we have another kind, we could update list of apicast kinds for tests that are supported.

testsuite/gateways/apicast/system.py
mastastny: If route exists, it's deleted and recreated. Brief downtime during delete+create, but resilient.Request retries cover it.

testsuite/gateways/apicast/system.py
mastastny: routes.create() is synchronous. It returns once the Route object is created in OCP. Route propagation may take a moment, but api_client uses resilient.Request which retries on 503/404, so that covers it.

testsuite/gateways/apicast/system.py
mastastny: Done — ZyncLessApicast(SystemApicast) is now a separate class in gateways/apicast/zyncless.py with lifecycle hooks. RHSSO works via ZYNC_OIDC_SYNC capability which is absent for ZyncLessApicast, so OIDC tests are automatically skipped.

doc/GATEWAYS.md
```diff
@@ -19,6 +21,35 @@ gateway:
   default:
     kind: "SystemApicast"
 ```
+## ZyncLess APIcast
+*Description*: Variant of System APIcast for 3scale deployments where zync is completely disabled. Since zync is not
+available to create OCP routes for APIcast endpoints, this gateway creates and manages them via lifecycle hooks.
+Requires `spec/zync/enabled: false` in the APIManager CR — capability provider raises an error if zync is enabled.
```
mdujava: other places uses `.` instead of `/` in CR paths

testsuite/capabilities/providers.py
```diff
@@ -62,3 +66,38 @@ def fips():
...
+    Provides ZYNC_ROUTES + ZYNC_OIDC_SYNC. Fails fast if APIManager has zync disabled.
+
+    Case 2 (zync without routes): threescale.zync_routes_disabled: true in config.
+    Provides only ZYNC_OIDC_SYNC.
+
+    Case 3 (zync disabled): gateway is ZyncLessApicast.
+    Provides no zync capabilities. Fails fast if APIManager has zync enabled.
+    """
+    if issubclass(gateways.default, ZyncLessApicast):
+        enabled = openshift().api_manager.get_path("spec/zync/enabled")
```
mdujava: enabled is used in this if, and outside no need to ask it in 2 places move it to the start

testsuite/capabilities/providers.py
```diff
@@ -62,3 +66,38 @@ def fips():
...
+        if enabled is Missing or enabled:
+            raise RuntimeError("ZyncLessApicast requires zync to be disabled in APIManager (spec/zync/enabled: false)")
+        return {}
+
+    zync_routes_disabled = weakget(settings)["threescale"]["zync_routes_disabled"] % False
+
+    if zync_routes_disabled:
+        return {Capability.ZYNC_OIDC_SYNC}
```
mdujava, line 111: we don't know here that zync is enabled and will do OIDC for us

testsuite/tests/conftest.py
```diff
@@ -99,6 +99,18 @@ def pytest_addoption(parser):
...
 
 
+def pytest_sessionstart(session: pytest.Session) -> None:
+    """Fail fast if zync configuration doesn't match the actual 3scale deployment.
+    Only runs on the controller process, not on xdist workers."""
+    if hasattr(session.config, "workerinput"):
+        return
+    try:
+        _ = Capability.ZYNC_ROUTES in CapabilityRegistry()
+        _ = Capability.ZYNC_OIDC_SYNC in CapabilityRegistry()
```
mdujava: do we need to call this 2 times? exception is thrown on CapabilityRegistry call, and we do not check result of `x in CapabilityRegistry()`

mdujava (changes requested):

testsuite/capabilities/providers.py
mastastny: Fixed, I also added check that  the route creation is disabled on the zync-que deployment.

testsuite/tests/conftest.py
mastastny: No we don't. Fixed.

testsuite/capabilities/providers.py
```diff
@@ -62,3 +66,51 @@ def fips():
...
+    zync_enabled = openshift().api_manager.get_path("spec/zync/enabled")
+
+    if issubclass(gateways.default, ZyncLessApicast):
+        if zync_enabled is Missing or zync_enabled:
+            raise RuntimeError("ZyncLessApicast requires zync to be disabled in APIManager (spec/zync/enabled: false)")
+        return {}
+
+    if zync_enabled is not Missing and not zync_enabled:
+        raise RuntimeError(
+            "APIManager has zync disabled. Use ZyncLessApicast gateway or set spec/zync/enabled: true in APIManager."
```
mdujava: again using `/` for APIManager path.

mdujava (dismissed):
in future please clear out the 'requested changes' so it does not show it if you addressed them

mdujava (approved):

mastastny:
> discussed required rebase and squash or combine commits to make change being in one commit instead of multiple, (e.g. adding zync capability and tagging the tests with it)

squashed

PR #1042, opened 2026-05-27
User permissions test

cathal-bailey:
- Aim is to expand scope of user permissions test
- Logic refactor to enable scope expansion
  - Views/Permissions defined in separate dictionaries 
  - Checks if user has required permission to access particular page
  - If true, expects page to be returned. Else, user access denied
- Broadening test scope: Adding `monitoring`, `plans` and `partners`
- Replacing `page.is_displayed` with less strict test. No longer checks for individual page elements which for some users don't appear, allowing for test scope to be expanded.

### DIFF
--- a/testsuite/tests/ui/users_and_roles/conftest.py
+++ b/testsuite/tests/ui/users_and_roles/conftest.py
@@ -1,0 +1,33 @@
+"""Conftest for user permissions tests"""
+
+import pytest
+from selenium.common.exceptions import NoSuchElementException, WebDriverException
+from widgetastic.widget import GenericLocatorWidget
+
+
+@pytest.fixture
+def is_page_accessible():
+    """
+    Helper function for permission tests to check if a page is accessible.
+
+    This checks for successful page access by verifying:
+    1. Correct URL path is loaded
+    2. The masthead header is present (exists on all allowed pages, not on access denied)
+    """
+
+    def _check(page):
+        if page.path not in page.browser.url:
+            return False
+
+        try:
+            masthead = GenericLocatorWidget(
+                page, locator="//header[contains(@class, 'pf-c-masthead') and contains(@class, 'pf-m-display-inline')]"
+            )
+            if not masthead.is_displayed:
+                return False
+        except (NoSuchElementException, WebDriverException):
+            return False
+
+        return True
+
+    return _check
--- a/testsuite/tests/ui/users_and_roles/test_permissions.py
+++ b/testsuite/tests/ui/users_and_roles/test_permissions.py
@@ -2,76 +2,68 @@
 
 import pytest
 
-from testsuite.ui.views.admin.audience.billing import BillingSettingsView, BillingView
+from testsuite.ui.views.admin.audience.account import AccountsView
+from testsuite.ui.views.admin.audience.billing import (
+    BillingSettingsView,
+    BillingView,
+)
 from testsuite.ui.views.admin.audience.developer_portal import (
     ActiveDocsView,
     CMSNewPageView,
     CMSNewSectionView,
     DeveloperPortalContentView,
 )
+from testsuite.ui.views.admin.backend.analytics import BackendTrafficView
 from testsuite.ui.views.admin.foundation import AccessDeniedView
 
-PERMISSION_DICT = [
-    pytest.param("portal", DeveloperPortalContentView),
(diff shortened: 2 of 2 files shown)


testsuite/tests/ui/users_and_roles/test_permissions.py
```diff
@@ -2,76 +2,58 @@
...
+    Tests user permissions permission per permission section
+        - Creates a member user with a specific permission
+        - Logs in as that member user
+        - Attempts to access a specific UI page
+        - If users permission matches page's required permission -> allowed
+        - Else, access denied
+        - partners, settings, monitoring and policy_registry to be added? More complicated
+    """
+    member_user = provider_member_user(allowed_sections=user_permission, allowed_services=False)
     custom_admin_login(member_user.entity_name, "123456")
```
mdujava: I would change it to use fixture `account_password`

cathal-bailey: fixed

mastastny:
Rebase please, it doesn't work anymore due to incompatibility with wait_for 2.0. The fix is already in the main.

cathal-bailey:
> Rebase please, it doesn't work anymore due to incompatibility with wait_for 2.0. The fix is already in the main.

done

testsuite/tests/ui/users_and_roles/test_permissions.py
```diff
@@ -2,76 +2,68 @@
...
-    pytest.param("plans", ActiveDocsView),
-]
-
-
-@pytest.fixture()
-def all_page_objects():
-    """Returns all page objects from permissions tuple filtered of views with same permission"""
-
-    def _all_page_objects(except_permission, current_view):
+PERMISSIONS = ["portal", "finance", "settings", "partners", "monitoring", "plans", "policy_registry"]
```
mdujava: no need to have permissions listed in 2 placesth

is's just `[mapping[0] for mapping in VIEWS]`

testsuite/tests/ui/users_and_roles/conftest.py
```diff
@@ -0,0 +1,33 @@
+"""Conftest for user permissions tests"""
+
+import pytest
+from selenium.common.exceptions import NoSuchElementException, WebDriverException
+from widgetastic.widget import GenericLocatorWidget
+
+
+@pytest.fixture
+def is_page_accessible():
```
mdujava: if this fixture is not being used in more tests, it can be defined directly in `test_*.py`. 
Also scope can be session or module. or be even strait up funcion (no deps on other fixtures) maby create `testsuite/ui/utils.py`

testsuite/tests/ui/users_and_roles/test_permissions.py
```diff
@@ -2,76 +2,68 @@
...
-        all_views = [
-            view
-            for perm, view in [param.values for param in PERMISSION_DICT]
-            if perm != except_permission or view == current_view
-        ]
-
-        return all_views
-
-    return _all_page_objects
+VIEWS = [
```
mastastny, line 21: Why it doesn't also test some views with `required_permission="settings"` and `required_permission="policy_registry"` ?

testsuite/tests/ui/users_and_roles/test_permissions.py
mastastny: It's not. In first values in the Views is subset of values from PERMISSIONS. Some values are there also multiple times.

mdujava: yeah, it looked from beginning..., you can still get a Set so its without duplicates, but it would require to have at least one view per each permission

testsuite/tests/ui/users_and_roles/test_permissions.py
cathal-bailey: If memory serves me I investigated adding full coverage for everything at the time, but some views are significantly more complicated than others. For example, `settings` page views require `page_id` argument, which the test doesn't currently have. `page_id` is dynamic and changes per test run, so it would require some thinking as to how we could add it. I think at the time I made the decision to expand the test to additional views where possible without a major refactoring. We can add increased coverage again, but would need to solve for these issues.

mastastny (approved):

mdujava:
@cathal-bailey please rebase feature branch instead of merging changes from main into it.

PR #1052, opened 2026-06-16
Add test for user-defined env var persistence in APIcast operator

cathal-bailey:
- Adding a new test to check that user-defined environment variables remain after operator reconciliation cycles. 
- This test validates fix for https://issues.redhat.com/browse/THREESCALE-12224
- APIcast operator update included in `2.16.4` GA
- Test forces APIcast operator to trigger reconciliation by changing a supported env var (`APICAST_LOG_LEVEL`)

### DIFF
--- a/testsuite/tests/apicast_operator/test_user_env_var.py
+++ b/testsuite/tests/apicast_operator/test_user_env_var.py
@@ -1,0 +1,58 @@
+"""
+Test that user defined environment variables persist through APIcast operator reconciliation.
+
+This test validates the fix for THREESCALE-12224.
+"""
+
+import pytest
+from packaging.version import Version
+
+from testsuite import TESTED_VERSION
+from testsuite.capabilities import Capability
+
+pytestmark = [
+    pytest.mark.sandbag,  # Test uses operator, is slow
+    pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-12224"),
+    pytest.mark.required_capabilities(Capability.OCP4),
+    pytest.mark.skipif(TESTED_VERSION < Version("2.16.4"), reason="TESTED_VERSION < Version('2.16.4')"),
+]
+
+
+@pytest.fixture(scope="module")
+def gw_with_user_env(staging_gateway, logger):
+    """
+    Adds environment variables to the APIcast deployment.
+    """
+
+    test_env_var_name = "APICAST_PATH_ROUTING_ONLY"
+    test_env_var_value = "true"
+
+    environ = staging_gateway.deployment.environ()
+    environ.refresh()
+
+    logger.info("Adding env var: %s=%s", test_env_var_name, test_env_var_value)
+    environ[test_env_var_name] = test_env_var_value
+
+    environ.refresh()
+    assert environ[test_env_var_name] == test_env_var_value, "Env var was not set correctly"
+
+    return environ, test_env_var_name, test_env_var_value
+
+
+def test_user_env_var_persists(gw_with_user_env, staging_gateway):
+    """Verify user defined vars remain after reconciliation"""
+    environ, test_env_var_name, test_env_var_value = gw_with_user_env
+
+    # Trigger reconciliation by modifying a supported env var that goes through the operator
+    staging_gateway.environ["APICAST_LOG_LEVEL"] = "debug"
+    staging_gateway.deployment.wait_for()
+
+    environ.refresh()
+
+    try:
+        current_value = environ[test_env_var_name]
+        assert (
+            current_value == test_env_var_value
+        ), f"Env var changed after reconciliation. Expected: {test_env_var_value}, got {current_value}"
+    except KeyError:
+        pytest.fail(f"FAILURE: User-defined env var '{test_env_var_name}' was removed during operator reconciliation.")


testsuite/tests/apicast_operator/test_user_env_var.py
```diff
@@ -0,0 +1,59 @@
...
+    environ = apicast_deployment.environ()
+    environ.refresh()
+
+    logger.info("Adding env var: %s=%s", test_env_var_name, test_env_var_value)
+    environ[test_env_var_name] = test_env_var_value
+
+    environ.refresh()
+    assert environ[test_env_var_name] == test_env_var_value, "Env var was not set correctly"
+
+    time.sleep(30)
```
mdujava: magic constant, there might be timestamp in CR when was object read/updated.

testsuite/tests/apicast_operator/test_user_env_var.py
```diff
@@ -0,0 +1,59 @@
...
+]
+
+
+@pytest.fixture(scope="module")
+def apicast_deployment(staging_gateway):
+    """
+    Returns the APIcast deployment object for env var manipulation
+    """
+    return staging_gateway.deployment
+
```
mdujava: this has has no complexity that needs to be in a separate function, using directly `staging_gateway.deployemnt` should be fine.

testsuite/tests/apicast_operator/test_user_env_var.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+from packaging.version import Version
+
+from testsuite import TESTED_VERSION
+
+pytestmark = [
+    pytest.mark.sandbag,  # Test uses operator, is slow
+    pytest.mark.issue("https://issues.redhat.com/browse/THREESCALE-12224"),
+    pytest.mark.nopersistence,  # Modifies operator managed infrastructure
```
mdujava: This should be fine to be tested in a upgrade tests. e.g create apicast CR in 2.16.4, and update to 2.16.x or 2.17 should still preserve env, but then setup of variable should be in a separate fixture (so test just checks that viable exists and its set to expected variable)

testsuite/tests/apicast_operator/test_user_env_var.py
```diff
@@ -0,0 +1,61 @@
...
+    staging_gateway.environ["APICAST_LOG_LEVEL"] = "debug"
+
+    return staging_gateway, test_env_var_name, test_env_var_value
+
+
+def test_user_env_var_persists(gw_with_user_env, logger):
+    """Verify user defined vars remain after reconciliation"""
+    staging_gateway, test_env_var_name, test_env_var_value = gw_with_user_env
+
+    logger.info("Checking if user-defined env var persists after reconciliation...")
```
mdujava: I think this log is unnecessary.

testsuite/tests/apicast_operator/test_user_env_var.py
```diff
@@ -0,0 +1,61 @@
...
+
+    logger.info("Adding env var: %s=%s", test_env_var_name, test_env_var_value)
+    environ[test_env_var_name] = test_env_var_value
+
+    environ.refresh()
+    assert environ[test_env_var_name] == test_env_var_value, "Env var was not set correctly"
+
+    # Trigger reconciliation by modifying a supported env var that goes through the operator
+    logger.info("Triggering operator reconciliation...")
+    staging_gateway.environ["APICAST_LOG_LEVEL"] = "debug"
```
mdujava: where is waiting on operator to reconcile.
if setup and test run fast enough it might happen before operator reconcile

maybe moving this to test would be better:
* setup prepares the env variable
* test will trigger reconciliation, waits on it, and checks env value

mdujava (changes requested):

testsuite/tests/apicast_operator/test_user_env_var.py
cathal-bailey: Apologies, put that in there during debug. I'll remove.

mdujava (dismissed):

testsuite/tests/apicast_operator/test_user_env_var.py
```diff
@@ -0,0 +1,58 @@
...
+    return environ, test_env_var_name, test_env_var_value
+
+
+def test_user_env_var_persists(gw_with_user_env, staging_gateway):
+    """Verify user defined vars remain after reconciliation"""
+    environ, test_env_var_name, test_env_var_value = gw_with_user_env
+
+    # Trigger reconciliation by modifying a supported env var that goes through the operator
+    staging_gateway.environ["APICAST_LOG_LEVEL"] = "debug"
+    staging_gateway.deployment.wait_for()
```
mdujava: wait_for is already done when `environ` is updated
but more important is that this should be marked as disruptive as it restarts staging apicast, but than it will not be running in persistence tests

cathal-bailey: I can remove the redundant wait_for. 
I don't think it is possible to properly validate that the env var persists without restarting staging apicast, so I guess it is better to mark it disruptive? WDYT?

cathal-bailey: `wait_for()` removed

mdujava:
commit message is too long
and before merging remove the "merge 'main'" and fixup commits

PR #1054, opened 2026-06-30
Fix for test_edit_tenant failure triggered by widgetastic fill() method

cathal-bailey:
- Widgetastic's `TextInput.fill()` method is appending text instead of replacing the existing value in the input field, even though it's supposed to clear and replace by default. This is causing `test_edit_tenant` to fail persistently.
- A `replace()` method has been added to tenant.py, alongside the existing `update()` method
- The `replace()` accessed the element directly through selenium and forces `self.org_name` to clear before updating it, which resolves the failure


mdujava:
Is this the only time `.fill(` does not replace the value? do we need original function then, I assume its never used, so I wouldn't like to keep unused function.

Maybe its worth to check if element is defined same way as other inputs (without this problem), if there are differences and maybe debounc might sneak some data after selenium fill, and react state combine them ...

cathal-bailey:
> Maybe its worth to check if element is defined same way as other inputs (without this problem), if there are differences and maybe debounc might sneak some data after selenium fill, and react state combine them ...

No other tests are using the `TenantEditView.update()`, so method separation is redundant. I'll fix this. 

There are some other tests using `fill()` method to edit existing fields, but the implementation is slightly different, like here for instance: 
https://github.com/3scale-qe/3scale-tests/blob/7ee51cc3824b47858d0cc9de68848e6e0603f4c8/testsuite/tests/ui/apiap/test_backend.py#L42

https://github.com/3scale-qe/3scale-tests/blob/7ee51cc3824b47858d0cc9de68848e6e0603f4c8/testsuite/ui/views/admin/backend/backend.py#L78

I don't believe we've seen any fails from the other tests yet. I suspect the failures in `test_edit_tenant` may be timing related (Form field not full ready when `fill()` is called), so I think it is better to explicitly clear the field first to ensure the test works correctly.

mdujava:
as its same fill followed by button click in both tests, I would suspect that tenant form webpage js might be doing something different;
first report to dev or check if we don't see some differences in a js => no need to "relax" test, but fix the issue

Also I have seen that there are 2 different strategies for fill, one of which waits on some kind of presence on webpage so maybe we just need to update which strategy to use

PR #1056, opened 2026-07-10
Fix persistence issue in test_app_id_special_chars

mastastny:
Split shared service into two (headers and query) so credentials_location is not mutated between parametrized tests. Fixes 403 on persistence load when headers test runs after query test overwrote the setting.

reproducer:
```pipenv run python -m pytest  -m 'not flaky' --fuzz --drop-nopersistence --store x  testsuite/tests/fuzz/test_app_id_special_chars.py```
```pipenv run python -m pytest  -m 'not flaky' --fuzz --drop-nopersistence --load x  testsuite/tests/fuzz/test_app_id_special_chars.py```

### DIFF
--- a/testsuite/tests/fuzz/test_app_id_special_chars.py
+++ b/testsuite/tests/fuzz/test_app_id_special_chars.py
@@ -23,7 +23,7 @@ def private_base_url(tools):
 
 @pytest.fixture(scope="module")
 def service_proxy_settings(service_proxy_settings):
-    "auth via url params doesn't work because of url encoding of special characters"
+    """auth via url params doesn't work because of url encoding of special characters"""
     service_proxy_settings.update(credentials_location="headers")
     return service_proxy_settings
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mdujava** [comment]
can't we put service_1 if _ else service_2 into the @pytest.parametrize?

**Comment by mastastny** [comment]
> can't we put service_1 if _ else service_2 into the @pytest.parametrize?

I don't think so. Parametrize values are resolved at collection time, before fixtures are instantiated, so we can't reference fixture objects there. But I might be missing something, do you have a specific approach in mind?

File: testsuite/tests/fuzz/test_app_id_special_chars.py
Comment (mdujava) [inline testsuite/tests/fuzz/test_app_id_special_chars.py]: if there is "headers" why not put here `header_service` fixture directly (`query_fixture` instead of "query"), similar how we in some tests use either prod_client or api_client.

Comment (mastastny) [reply testsuite/tests/fuzz/test_app_id_special_chars.py]: Good point, but credentials_location is also used in the assertion (line 236) to decide whether to check headers or params, so we'd still need it as a value, not just a fixture reference.

File: testsuite/tests/fuzz/test_app_id_special_chars.py [hunk 1]
```diff
@@ -36,27 +36,42 @@ def service_settings(service_settings):
...
+    """Reuse application plan for all params applications"""
+    return custom_app_plan(rawobj.ApplicationPlan(blame(request, "aplan")), service_params)
+
+
+# pylint: disable=too-many-arguments
+@pytest.fixture()
+def application_headers(app_id, app_key, custom_application, app_plan_headers, lifecycle_hooks, request):
+    """Application with credentials sent via headers"""
+    app_obj = rawobj.Application(blame(request, "app"), app_plan_headers, app_id=app_id, app_key=app_key)
+    return custom_application(app_obj, hooks=lifecycle_hooks)
```
mdujava: last nitpick can you move this up, just bellow app_plan_headers, so _headers and _params are next to each other?

mdujava:
added few changes into there you can do `git rebase HEAD~2 --autosquash` and rebase please

mdujava (approved):

dhlavac (approved):

PR #1075, opened 2026-08-05
ci: add stable_create_backport_prs workflow

mdujava:
Adding new workflow that will create backporting PRs into stable-* branches when PR with `backport-to-2.X` is merged

### DIFF
--- a/.github/workflows/stable_create_backport_prs.yaml
+++ b/.github/workflows/stable_create_backport_prs.yaml
@@ -1,0 +1,156 @@
+name: stable_create_backport_prs
+
+# Trigger: a PR targeting main is merged AND carries the backport-to-X.Y label.
+# For each matching label the workflow cherry-picks the squash/merge commit onto
+# the corresponding stable-X.Y branch and opens a backport PR.
+# The label format must be exactly "backport-to-X.Y" (e.g. backport-to-2.15).
+
+on:
+  pull_request:
+    types: [closed, labeled]
+    branches:
+      - main
+
+jobs:
+  stable_create_backport_prs:
+    if: >
+      github.event.pull_request.merged == true &&
+      (
+        (github.event.action == 'closed' && contains(toJson(github.event.pull_request.labels.*.name), 'backport-to-')) ||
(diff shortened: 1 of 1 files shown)


.github/workflows/stable_create_backport_prs.yaml
```diff
@@ -0,0 +1,156 @@
...
+        if: steps.check.outputs.do_backport == 'true' && steps.push.outputs.conflict == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.BOTBEZBOT_TOKEN }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          TARGET: ${{ steps.check.outputs.target_branch }}
+          MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
+          PR_AUTHOR: ${{ github.event.pull_request.user.login }}
+        run: |
+          BACKPORT_BRANCH="backport/pr-${PR_NUMBER}-to-${TARGET}"
+          TITLE=$(gh pr view "$PR_NUMBER" --repo 3scale-qe/3scale-tests --json title --jq .title)
```
dhlavac, line 115: maybe better --repo "$GH_REPO" - to use possibly on forks ?

dhlavac:
just one thing otherwise LGTM! (only thing what i potentialy see is sha based cheripicking - if we manualy cheripick there will be conflict - but handled later in PR  - possible creating a patch and application of the patch to backported branch is also way but LGTM as it is)

.github/workflows/stable_create_backport_prs.yaml
mdujava: I do not want to automate use or create PR on a forks

mdujava:
> just one thing otherwise LGTM! (only thing what i potentialy see is sha based cheripicking - if we manualy cheripick there will be conflict - but handled later in PR - possible creating a patch and application of the patch to backported branch is also way but LGTM as it is)

if there is conflict, you get comment on original PR, how to make backport manualy

dhlavac (approved):

cathal-bailey (approved):

